PR #1, opened 2021-12-26
Add basic test

rikhuijzer:
Adds tests via a workflow in this repository.


rikhuijzer:
Oops PR against wrong repo. This is still a work in progress.

rikhuijzer:
The test almost works: https://github.com/rikhuijzer/cache-artifacts/pull/2. I'll fix it within a few days.

SaschaMann:
Sounds great, please add yourself to the authors as well since you've already done more work than I have for this action :D

action.yml
```diff
@@ -18,17 +20,18 @@ outputs:
...
         key: ${{ runner.os }}-test-${{ inputs.cache-name }}-${{ hashFiles('**/Project.toml') }}
         restore-keys: |
           ${{ runner.os }}-test-${{ inputs.cache-name }}-
-          ${{ runner.os }}-test-
-          ${{ runner.os }}-
 
-    - run: echo "::set-output name=cache-hit::$CACHE_HIT"
+    - run: |
+        echo "Setting outputs.cache-hit to \"$CACHE_HIT\""
+        echo "::set-output name=cache-hit::$CACHE_HIT"
```
rikhuijzer, line 34: @SaschaMann do you have an idea why this is empty? I've tried switching to another actions/cache version but that didn't solve it. It's very weird because it should definitely work. The failure of this step causes `test-output` and `test-restore` to fail.
``` 
Run echo "Setting outputs.cache-hit to \"$CACHE_HIT\""
Setting outputs.cache-hit to ""
```
(https://github.com/rikhuijzer/cache-artifacts/actions/runs/1626657972).

rikhuijzer: Got it! cache-hit is "" when the restore-key doesn't exist and "false" when it is a cache miss

rikhuijzer:
All tests pass now :partying_face: (see https://github.com/rikhuijzer/cache-artifacts/pull/2)

Some notes:

- The tests are quite extensive now. I guess that that's good because it would be very unfortunate it if turns out that this action doesn't really cache artifacts properly. So, the tests cover outputs.cache-hit AND the actual artifacts folders
- To ensure that each run stores a new cache, I've had to make the restore-keys more strict. The original idea of those keys was probably that the Artifacts system can handle files for different versions perfectly fine, so more files is better. I wonder now in what kind of situation these stricter restore-keys make a difference. I think it should be fine
- @SaschaMann could you enable GitHub Actions on this repository?

SaschaMann:
> To ensure that each run stores a new cache, I've had to make the restore-keys more strict. The original idea of those keys was probably that the Artifacts system can handle files for different versions perfectly fine, so more files is better. I wonder now in what kind of situation these stricter restore-keys make a difference. I think it should be fine

I think originally they were based on the example GitHub provided rather than Julia-specific. Should be fine with stricter keys too.

> @SaschaMann could you enable GitHub Actions on this repository?

Done

.github/workflows/CI.yml
```diff
@@ -0,0 +1,69 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - "README.md"
+      - "LICENSE"
```
SaschaMann, line 9:
```suggestion
    paths:
      - 'action.yml'
```

.github/workflows/CI.yml
```diff
@@ -0,0 +1,69 @@
...
+  pull_request:
+    paths-ignore:
+      - "README.md"
+      - "LICENSE"
```
SaschaMann, line 13:
```suggestion
    paths:
      - 'action.yml'
```

.github/workflows/CI.yml
```diff
@@ -0,0 +1,69 @@
...
+
+jobs:
+  test-save:
+    runs-on: ubuntu-latest
+    outputs:
+      cache-name: ${{ steps.name.outputs.cache-name }}
+    steps:
+    - uses: actions/checkout@v2
```
SaschaMann, line 21:
```suggestion
    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
```

I prefer pinning actions to commit SHAs for security reasons. We can have Dependabot update them if needed.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,69 @@
...
+      run: 'write("random.txt", string(rand(10)))'
+    - name: Save cache
+      id: cache
+      uses: ./
+      with:
+        cache-name: ${{ hashFiles('random.txt') }}
+    - name: Set cache-name as output
+      id: name
+      run: |
+        echo "Setting outputs.cache-name to \"$CACHE_NAME\""
```
SaschaMann, line 33: When debug logs are enabled, this will be displayed anyway. I don't mind the duplication though.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,69 @@
...
+        @assert !isdir(dir)
+    - name: Install a small binary
+      shell: 'julia --color=yes {0}'
+      run: 'using Pkg; Pkg.add("pandoc_jll")'
+
+  test-restore:
+    needs: test-save
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
SaschaMann, line 50:
```suggestion
    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
```

action.yml
```diff
@@ -1,6 +1,8 @@
 name: 'Cache Julia Artifacts'
 description: 'Cache Julia artifacts using actions/cache'
-author: 'Sascha Mann'
+author:
+  - 'Sascha Mann'
+  - 'Rik Huijzer'
```
SaschaMann, line 5: Does this accept a list rather than a string (with comma separated names or so)? I've not tried it and the docs aren't explicit about it: https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#author

action.yml
```diff
@@ -25,10 +28,11 @@ runs:
         key: ${{ runner.os }}-test-${{ inputs.cache-name }}-${{ hashFiles('**/Project.toml') }}
         restore-keys: |
           ${{ runner.os }}-test-${{ inputs.cache-name }}-
-          ${{ runner.os }}-test-
-          ${{ runner.os }}-
 
-    - run: echo "::set-output name=cache-hit::$CACHE_HIT"
+    - id: hit
+      run: |
+        echo "Setting outputs.cache-hit to \"$CACHE_HIT\""
```
SaschaMann, line 34: Since this runs in the user's repos, I think we should avoid the duplication here. The debug logs already show that outputs have been set.

```suggestion
```

SaschaMann (commented):
Looks good, some questions and minor comments.

Thanks for creating these :)

SaschaMann:
Apparently closing and reopening doesn't trigger CI anymore, could you push an empty commit or so to trigger it?

action.yml
rikhuijzer: I have no idea. I cannot see it even when I try to publish to the GitHub Marketplace.

Let's merge like this and then change it back if it doesn't work? We could also change the author to `julia-actions`

rikhuijzer: I've changed the author to `julia-actions`. Let me know if you want anything different. It's not so important for me what's there.

action.yml
```diff
@@ -9,26 +10,29 @@ branding:
...
   using: 'composite'
   steps:
     - uses: actions/cache@937d24475381cd9c75ae6db12cb4e79714b926ed
       id: cache
       with:
-        path: ~/.julia/artifacts
+        path: |
+          ~/.julia/artifacts
+          ~/.julia/packages
+          ~/.julia/registries
```
giordano, line 29: To be clear: I do _**not**_ recommend always caching all these directories, especially the registry on Windows with Julia v1.6- can actually cause a slowdown rather than a speedup.

rikhuijzer:
> Apparently closing and reopening doesn't trigger CI anymore, could you push an empty commit or so to trigger it?

Doesn't seem to work. I guess we first need to get this `.github/workflow/` into main?

Tests passed again for a6f12e0, see https://github.com/rikhuijzer/cache-artifacts/pull/4.

action.yml
rikhuijzer: Okay. Then, I'll leave that for a different PR.

SaschaMann (approved):

PR #4, opened 2021-12-27
Extend cache action

### LINKED ISSUE #2: Extend the action to packages and registry
In some cases it may be beneficial to cache also packages (`.julia/packages`) and/or the registry (`.julia/registries`), not just artifacts.  For example I cache the registry in [`Measurements.jl`](https://github.com/JuliaPhysics/Measurements.jl/blob/b7b7c8d7ac31bcfd51088dcd9a18c0574b6d368c/.github/workflows/ci.yml#L40), and today I [enabled caching of packages](https://github.com/JuliaRegistries/General/pull/51270) in CI for the General registry.  Note that this would affect also the name of the repository/action.

Maybe we can have options to enable cachin of artifacts, packages and registy, with artifacts being the only one enabled by default?

rikhuijzer:
Closes #2.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -38,7 +38,7 @@ runs:
           echo "At least one of cache-artifacts, cache-packages or cache-registry has to be true"
           exit 1
         fi
-        echo "::set-output name=cache-paths::$(echo "$PATHS")"
+        echo "::set-output name=cache-paths::$PATHS"
       env:
         CACHE_ARTIFACTS: ${{ inputs.cache-artifacts }}
         CACHE_PACKAGES: ${{ inputs.cache-packages }}
--- a/action.yml
+++ b/action.yml
@@ -35,7 +35,7 @@ runs:
         [ "$CACHE_PACKAGES" = "true" ] && PATHS="${PATHS}~/.julia/packages"$'\n'
         [ "$CACHE_REGISTRY" = "true" ] && PATHS="${PATHS}~/.julia/registry"$'\n'
         if [ "$PATHS" = "" ]; then
-          echo "At least one of cache-artifacts, cache-packages or cache-registry has to be true"
+          echo "::error title=Invalid inputs::At least one of cache-artifacts, cache-packages or cache-registry has to be true"
           exit 1
         fi
         echo "::set-output name=cache-paths::$PATHS"
--- a/action.yml
+++ b/action.yml
@@ -1,5 +1,5 @@
-name: 'Cache Julia Artifacts'
-description: 'Cache Julia artifacts using actions/cache'
+name: 'Cache Julia artifacts, packages and registry'
+description: 'Cache Julia using actions/cache'
 author: 'Sascha Mann, Rik Huijzer, and contributors'
 
 branding:
@@ -9,7 +9,16 @@ branding:
 inputs:
   cache-name:
     description: 'Name used as part of the cache keys'
-    default: 'cache-artifacts'
+    default: 'julia-cache'
+  cache-artifacts:
+    description: 'Whether to cache artifacts'
+    default: 'true'
+  cache-packages:
+    description: 'Whether to cache packages'
+    default: 'true'
+  cache-registry:
+    description: 'Whether to cache the registry'
+    default: 'false'
 
 outputs:
   cache-hit:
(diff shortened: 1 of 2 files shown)


rikhuijzer:
Very weird. Using \n as a path separator and empty lines should work fine:

https://github.com/actions/cache/blob/4498c5b4d8b0673dbc7d9731bc11d9cc5a2c6087/__tests__/actionUtils.test.ts#L201-L218

rikhuijzer:
The current failures are because the cache-name which is passed from `test-save` to `test-restore` is not always passed to the same operating system. Then, the restore key doesn't match anymore so it is an unexpected cache miss.

rikhuijzer:
Great. Everything works again and is now tested on all three operating systems.

I've considered testing the PATHS logic more, but chose not too for now. It is already covered pretty well since currently two things are tested:
- artifacts is inside PATHS (otherwise, the artifacts will not be cached and the last job in CI.yml will fail)
- the PATHS newlines are properly propagated into `cache.key` (otherwise, nothing will be cached at all and the last job in CI.yml will fail)

action.yml
```diff
@@ -19,11 +28,27 @@ outputs:
...
+      run: |
+        PATHS=""
+        [ "$CACHE_ARTIFACTS" = "true" ] && PATHS="${PATHS}~/.julia/artifacts"$'\n'
+        [ "$CACHE_PACKAGES" = "true" ] && PATHS="${PATHS}~/.julia/packages"$'\n'
+        [ "$CACHE_REGISTRY" = "true" ] && PATHS="${PATHS}~/.julia/registry"$'\n'
+        if [ "$PATHS" = "" ]; then
+          echo "At least one of cache-artifacts, cache-packages or cache-registry has to be true"
+          exit 1
+        fi
+        echo "::set-output name=cache-paths::$(echo "$PATHS")"
```
SaschaMann, line 41:
```suggestion
        echo "::set-output name=cache-paths::$PATHS"
```

action.yml
```diff
@@ -19,11 +28,27 @@ outputs:
...
   using: 'composite'
   steps:
+    - id: paths
+      run: |
+        PATHS=""
+        [ "$CACHE_ARTIFACTS" = "true" ] && PATHS="${PATHS}~/.julia/artifacts"$'\n'
+        [ "$CACHE_PACKAGES" = "true" ] && PATHS="${PATHS}~/.julia/packages"$'\n'
+        [ "$CACHE_REGISTRY" = "true" ] && PATHS="${PATHS}~/.julia/registry"$'\n'
+        if [ "$PATHS" = "" ]; then
+          echo "At least one of cache-artifacts, cache-packages or cache-registry has to be true"
```
SaschaMann, line 38:
```suggestion
          echo "::error title=Invalid inputs::At least one of cache-artifacts, cache-packages or cache-registry has to be true"
```

SaschaMann (commented):
Looks good to me, could you update the README as well, please? Should we rename this to julia-actions/cache then?

rikhuijzer:
> Should we rename this to julia-actions/cache then?

Maybe, `cache-julia` to be in line with `setup-julia`, `julia-buildpkg`, `julia-runtest` etc? I find the names a bit double so just `cache` makes more sense indeed. Maybe to avoid clashing fork names?

Good idea to rename. I'll leave the decision up to you to what exactly. I'm okay with `cache`, `cache-julia` and similar names.

SaschaMann:
`cache-julia` sounds like it caches Julia which it doesn't (and probably shouldn't), I'll go with just `cache` then

rikhuijzer:
> `cache-julia` sounds like it caches Julia which it doesn't (and probably shouldn't), I'll go with just `cache` then

Fair enough :+1:

I'm seeing some possible discussion points in the README too. Can we merge this first to avoid having a too long PR? Then, I'll open a new PR for the README

SaschaMann:
> I'm seeing some possible discussion points in the README too. Can we merge this first to avoid having a too long PR? Then, I'll open a new PR for the README

Yea that's fine. I'll tag a new release once the README has been updated then :)

rikhuijzer:
Thank you for that and also for the reviews :smile: I've learned some things from them

I'll do the README tomorrow.

PR #5, opened 2021-12-28
Also test packages dir

rikhuijzer:
I found a bug. Only the artifacts dir was cached even when more paths were passed in. So, I expanded the tests and confirmed that the test failed. The problem was caused by GitHub dropping newlines somewhere between setting the variable in the first step and reading the variable in the second step.

This is now solved by passing separate variables and joining them (`format`) in the cache step.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -69,10 +69,11 @@ jobs:
         @assert ENV["cache-hit"] == "true"
       env:
         cache-hit: ${{ steps.cache.outputs.cache-hit }}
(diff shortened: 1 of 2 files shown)


rikhuijzer:
To double check. I've tested this on another repository too. With `cache-registries: "true"`, the logs clearly show that many files in all three directories are getting restored. This should save a lot of time because GitHub puts all files in one neat zipped blob. That seems to be the case, because the running time is reduced by 97 - 69 = 28 seconds on the second run.

SaschaMann (approved):
Thank you!

PR #32, opened 2022-10-14
Remove deprecated set-output command

SaschaMann:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ & https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -47,7 +47,7 @@ runs:
           ${{ runner.os }}-test-${{ inputs.cache-name }}-
 
     - id: hit
-      run: echo "cache-hit$CACHE_HIT" >> $GITHUB_OUTPUT
+      run: echo "cache-hit=$CACHE_HIT" >> $GITHUB_OUTPUT
       env:
         CACHE_HIT: ${{ steps.cache.outputs.cache-hit }}
       shell: bash
--- a/action.yml
+++ b/action.yml
@@ -47,7 +47,7 @@ runs:
           ${{ runner.os }}-test-${{ inputs.cache-name }}-
 
     - id: hit
-      run: echo "::set-output name=cache-hit::$(echo $CACHE_HIT)"
+      run: echo "cache-hit$CACHE_HIT" >> $GITHUB_OUTPUT
       env:
         CACHE_HIT: ${{ steps.cache.outputs.cache-hit }}
       shell: bash


action.yml
```diff
@@ -47,7 +47,7 @@ runs:
           ${{ runner.os }}-test-${{ inputs.cache-name }}-
 
     - id: hit
-      run: echo "::set-output name=cache-hit::$(echo $CACHE_HIT)"
+      run: echo "cache-hit$CACHE_HIT" >> $GITHUB_OUTPUT
```
rikhuijzer, line 50:
```suggestion
      run: echo "cache-hit=$CACHE_HIT" >> $GITHUB_OUTPUT
```

giordano:
I can't answer the question, but honestly, the syntax of `set-output` was awful, I'm glad they changed it.  This is now a plain `var=value` list in a dedicated file, easier to read and parse than grabbing some specific text in the log.

SaschaMann:
When they moved `add-path` and `set-env`, it was due to a security issue: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

The explanation given in the blog post sounds similar

> **To avoid untrusted logged data** to use set-stateand set-output workflow commands without the intention of the workflow author we have introduced a [new set of environment files](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files) to manage state and output.

I'm guessing the whole idea of printing things to stdout as a way to run commands didn't work out because it's very hard to sanitize everything that's being printed in a CI log on potentially untrusted PRs and such?

Fully agree that the syntax was awful and this is much better, too.

rikhuijzer (approved):
Probably best to also remove the `set-output` at

https://github.com/julia-actions/cache/blob/d6968100711335869b395447423d3546516738fb/.github/workflows/CI.yml#L26

Apart from that the API remains the same, right? For the end-user. If yes and tests pass then all looks good to me :+1:

SaschaMann:
Ah good call, yes that should be updated as well. I only checked `action.yml`

Do you want to open a PR?

> Apart from that the API remains the same, right? For the end-user. If yes and tests pass then all looks good to me 👍

Yes, I think so. Unfortunately we'll have to tag the 2nd release today and spam people who use Dependabot a bit because I had already tagged a release for #31

rikhuijzer:
> Do you want to open a PR?

Easier if you do it, merge and tag all in one go I think.

Issue #11, opened 2022-02-13
Cache precompilation cache


giordano:
It may also be useful caching `.julia/compiled`, to avoid re-precompiling packages, whose sources are already cached by default.

rikhuijzer:
Yes I was also thinking about this. Could save minutes on some jobs with many dependencies.

Do you know whether this folder can be re-used between different CPUs?

giordano:
Uhm, there may be a problem for multi-architecture builds for same operating system: the slug [depends, among other things](https://github.com/JuliaLang/julia/blob/befe38f18641d0e49c3de81ae9ca21d1941cfd95/base/loading.jl#L1460-L1463) on the full absolute path of the sysimage and the julia executable, which would be different for different operating systems, but for example for x86 and x64 builds of Linux those paths might be the same (not 100% sure, you'd need to check)

giordano:
In the worst case this can be disabled by default and the README would explain clearly that the precompilation cache can be used only if you don't have multi-architecture jobs for the same operating system.

DilumAluthge:
Can we modify Julia to include the architecture when naming the precompilation files?

giordano:
That would mean modifying [`Base.compilecache_path`](https://github.com/JuliaLang/julia/blob/befe38f18641d0e49c3de81ae9ca21d1941cfd95/base/loading.jl#L1453-L1467).  It may be a good idea for the future, but won't apply to past versions even if accepted.

The easiest thing would be controlling where [`julia-actions/setup-julia`](https://github.com/julia-actions/setup-julia) installs Julia.  I have no idea if that includes the architecture somewhere.

giordano:
> The easiest thing would be controlling where [julia-actions/setup-julia](https://github.com/julia-actions/setup-julia) installs Julia. I have no idea if that includes the architecture somewhere.

Looks like it does: https://github.com/julia-actions/setup-julia/blob/2d6959702913c06ee8c6997dadafe089ec18549f/src/setup-julia.ts#L62?

SaschaMann:
Two things I'd be concerned about, without knowing details about what's stored in .julia/compiled and about how wrong files in that directory would affect things:

- How will partial cache hits affect this, e.g. when some packages were cached but others weren't? Or one cache includes an older version? Does Pkg figure things out on its own in that case, too?
- How will this work when several Julia versions are installed and used alongside each other in the same GHA environment? This could also be considered out of scope for this action fwiw.

[...]

PR #38, opened 2022-12-24
Add option to cache precompilation cache

SaschaMann:
See #11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ That is why caching the registries is disabled by default.
 - `cache-artifacts` - Whether to cache `~/.julia/artifacts/`. Enabled by default.
 - `cache-packages` - Whether to cache `~/.julia/packages/`. Enabled by default.
 - `cache-registries` - Whether to cache `~/.julia/registries/`. Disabled by default.
+- `cache-precompilation-cache` - Whether to cache `~/.julia/compiled/`. Disabled by default. **USE ONLY IF YOU KNOW WHAT YOU'RE DOING!** See [#11](https://github.com/julia-actions/cache/issues/11).
 
 ### Outputs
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -40,6 +40,7 @@ That is why caching the registries is disabled by default.
 - `cache-artifacts` - Whether to cache `~/.julia/artifacts/`. Enabled by default.
 - `cache-packages` - Whether to cache `~/.julia/packages/`. Enabled by default.
 - `cache-registries` - Whether to cache `~/.julia/registries/`. Disabled by default.
+- `cache-precompilation-cache` - Whether to cache `~/.julia/compiled/`. Disabled by default. **USE ONLY IF YOU KNOW WHAT YOU'RE DOING!** See [#11](https://github.com/julia-actions/cache/issues/11).
```
rikhuijzer, line 43: Should this be `cache-precompilation` to be more consistent with the other input names? The rest of the directories, such as `~/.julia/packages` are basically also caches.

Should the counterarguments you gave be put into the README? That seems like a more reliable/permanent location for the information.

SaschaMann: > Should this be cache-precompilation to be more consistent with the other input names? The rest of the directories, such as ~/.julia/packages are basically also caches.

Good idea. Maybe `cache-compiled` to make it match the directory name like the others do? 

> Should the counterarguments you gave be put into the README? That seems like a more reliable/permanent location for the information.

I considered it but the issue still contains a number of unknowns. Linking to it seemed like a better way to ensure people see the latest info. If/when those are resolved, we can add them to the README directly, I think.

giordano (approved):
Looks good to me as far as I can tell (I'm not really expert about this) and I'm happy with all the caution warnings.  Thanks!

PR #50, opened 2023-01-14
Use step outputs instead of env vars

### LINKED ISSUE #48: Use outputs instead of env vars
Seems like a better practice and may prevent issues with env var weirdness (#42)

SaschaMann:
fixes #48


IanButterworth (approved):

IanButterworth:
LGTM. I guess my review doesn't count on auto merge

giordano (approved):

Issue #7, opened 2021-12-28
Set `cache-registries` to true by default on non-Windows operating systems


rikhuijzer:
This was partially discussed in #2.

Currently, the `cache-registries` input is `false` by default because it may actually slow down running time on Windows. Many people will probably not notice this input and leave it on `false` or can't be bothered to ensure that this input is only set to `true` on non-Windows operating systems.

So, should we check whether the user didn't set anything and when the operating system is Ubuntu or macOS, then set `cache-registries: true`?

giordano:
The problem with Windows is that the `julia-buildpkg` and `julia-runtest` workflows unconditionally set `JULIA_PKG_SERVER=""`, so that the registry is _always_ a bloated 200+ MB git repositories with 30k+ small files, which is horribly slow for Windows.  By caching that, you put that in a tarball, but Windows is also horribly slow at unpacking tarballs with lots of small files, so that caching the registry in practice saves you nothing on that platform.  The situation is much different on Unix systems, where tarballs are handled just fine.

If those workflows didn't set `JULIA_PKG_SERVER=""`, caching the registry would make a bit more on all platforms, especially with Julia v1.7+ where the registry is just another tarball, although the size of the registry in this case would be of the order of 4 MB.

rikhuijzer:
Thanks again for the clarification

> If those workflows didn't set `JULIA_PKG_SERVER=""`, caching the registry would make a bit more on all platforms, especially with Julia v1.7+ where the registry is just another tarball, although the size of the registry in this case would be of the order of 4 MB.

In a test workflow that I ran, restoring the artifacts, packages and registries directories took 6 seconds on an Ubuntu Julia 1.7 runner. Putting all this stuff into the cache took 8 seconds on an earlier Ubuntu run. So, that's pretty quick. For now, however, it seems the safest bet indeed to not cache the registries by default. I'll probably do some benchmarks at a later point to see what's better where.

giordano:
https://github.com/JuliaPhysics/Measurements.jl/runs/4650042302?check_suite_focus=true#step:4:1 1m 22s on Windows, 15 s on macOS, 4s on Linux

IanButterworth:
I think this can be enabled now. 

https://github.com/julia-actions/julia-buildpkg/pull/31 - landed properly on https://github.com/julia-actions/julia-buildpkg/releases/tag/v1.4.2

[...]

PR #62, opened 2023-07-25
enable registries by default

IanButterworth:
Closes https://github.com/julia-actions/cache/issues/7

`julia-buildpkg` and `julia-runtest` now both use the pkgserver registry, so this should work

https://github.com/julia-actions/julia-buildpkg/pull/31 - landed properly on https://github.com/julia-actions/julia-buildpkg/releases/tag/v1.4.2

https://github.com/julia-actions/julia-runtest/pull/74 - landed properly on https://github.com/julia-actions/julia-runtest/releases/tag/v1.9.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,24 +22,14 @@ jobs:
     - uses: julia-actions/julia-runtest@v1
 ```
 
-By default, this caches the files in `~/.julia/artifacts/` and `~/.julia/packages/`.
-To also cache `~/.julia/registries/`, use
-
-```yaml
-    - uses: julia-actions/cache@v1
-      with:
-        cache-registries: "true"
-```
-
-Note that caching the registries may actually slow down the workflow running time on Windows runners.
-That is why caching the registries is disabled by default.
+By default, this caches the files in `~/.julia/artifacts/`,  `~/.julia/packages/` and `~/.julia/registries`.
 
 ### Inputs
 
 - `cache-name` - Name used as part of the cache keys
 - `cache-artifacts` - Whether to cache `~/.julia/artifacts/`. Enabled by default.
 - `cache-packages` - Whether to cache `~/.julia/packages/`. Enabled by default.
-- `cache-registries` - Whether to cache `~/.julia/registries/`. Disabled by default.
+- `cache-registries` - Whether to cache `~/.julia/registries/`. Enabled by default.
 - `cache-compiled` - Whether to cache `~/.julia/compiled/`. Disabled by default. **USE ONLY IF YOU KNOW WHAT YOU'RE DOING!** See [#11](https://github.com/julia-actions/cache/issues/11).
 
 ### Outputs
--- a/action.yml
+++ b/action.yml
@@ -18,7 +18,7 @@ inputs:
     default: 'true'
   cache-registries:
     description: 'Whether to cache ~/.julia/registries/'
-    default: 'false'
+    default: 'true'
   cache-compiled:
     description: 'Whether to cache ~/.julia/compiled. USE WITH CAUTION! See https://github.com/julia-actions/cache/issues/11 for caveats.'
     default: 'false'


IanButterworth:
@DilumAluthge can you review this please

DilumAluthge:
> `julia-buildpkg` and `julia-runtest` now both use the pkgserver registry, so this should work

Specifically, it's because they use the non-extracted tarball, right? IIRC, the problem with slowness on Windows was due to the time it takes to extract all the files?

DilumAluthge (approved):
Seems fine to me, and I think we can consider this a bug fix. Would be good to get review from someone more familiar than me with this action, though.

DilumAluthge:
> and I think we can consider this a bug fix

To elaborate what I meant: if we can consider this to be a bugfix, then we don't have to make a breaking release of this action.

rikhuijzer (approved):
I‘ve tried this a while ago on Linux runners and it worked fine and saved some time if I remember correctly.

I’m on vacation and have no computer at hand. Would be good if someone could do some benchmarks for this PR on, for example, latest Julia with Linux, Mac, and Windows. Maybe it takes longer on average and we probably wouldn’t quickly notice after merge.

SaschaMann:
+1 on considering this a bug fix. I don't have the capacity to benchmark it either but feel free to merge and tag a release if you deem it ready :)

IanButterworth:
I've been thinking and I'm not sure we should be caching the registries by default.

[Since 1.9, julia will only update the registry once per day,](https://julialang.org/blog/2023/04/julia-1.9-highlights/#pkg_add_will_only_auto_update_the_registry_once_per_day) unless an explicit `Pkg.update` is called.

If we cache the registry, and now that scratchspaces are cached by default (where Pkg stores the date of the last registry update), Pkg will be able to know when the registry was last updated and avoid doing it unless `update` is called.

That means that if a user wants to pick up a new version of a package that was just registered, simply re-running CI won't do that when the packages are `Pkg.add`-ed or `Pkg.instantiate`-ed for 24 hrs.

They'd be forced instead to either clear out the cache, or add a `Pkg.update`

Also not caching the registry won't hurt the goal of making affect precompile cache hits.

giordano:
> [Since 1.9, julia will only update the registry once per day,](https://julialang.org/blog/2023/04/julia-1.9-highlights/#pkg_add_will_only_auto_update_the_registry_once_per_day) unless an explicit Pkg.update is called.

The other actions could force a `Pkg.Registry.update` before installing the dependencies?

giordano:
However [I maintain](https://github.com/julia-actions/cache/issues/7#issuecomment-1002070272) that enabling caching of the git registry on Windows can be a performance hit (but the git registry on Windows is a problem anyway).

IanButterworth:
> The other actions could force a Pkg.Registry.update before installing the dependencies?

So the registry would be first downloaded from the cache, then an update downloaded. I think that defeats the point.

And yeah, I doubt we'd want to do this for git caches.

Perhaps we kick this ball further down the road.. The main gains to be had in duration are in getting precompile caches cached. Once that's stable we could see if this is worth it?

giordano:
> So the registry would be first downloaded from the cache, then an update downloaded. I think that defeats the point.

Not really: if you have a git registry, pulling a few commits is much quicker than doing a full clone, and if the compressed registry is already up-to-date nothing is downloaded at all, no?

omus:
So `Pkg.instantiate` will [automatically update the registry](https://github.com/JuliaLang/Pkg.jl/blob/3c86ba27e904807e13beb8cb0466ed70365b0b2d/src/API.jl#L1818-L1824) when no manifest file is present.

In the case where a manifest file is present then we require exact versions of packages so as long as the cached registry includes those required versions the instantiation should succeed. In a scenario where we're re-using a long lived cache and the manifest uses a version outside of the registry we end up [falling back to updating the registries](https://github.com/JuliaLang/Pkg.jl/blob/3c86ba27e904807e13beb8cb0466ed70365b0b2d/src/API.jl#L1850-L1859).

A quick example of `julia-actions/julia-runtest` updating the registries after using `cache-registries: true` with a successful cache hit:
```
Run julia-actions/julia-runtest@v1
Run echo "JULIA_PKG_SERVER_REGISTRY_PREFERENCE=${JULIA_PKG_SERVER_REGISTRY_PREFERENCE:-eager}" >> ${GITHUB_ENV}
Run # Functionality only currently works on a narrow range of Julia versions... see #76
Run # The Julia command that will be executed
    ...
    Updating registry at `~/.julia/registries/General`
    Updating git-repo `git@github.com:JuliaRegistries/General.git`
```

I think we're good to enable registry caching by default.

README.md
```diff
@@ -22,24 +22,14 @@ jobs:
...
-
-```yaml
-    - uses: julia-actions/cache@v1
-      with:
-        cache-registries: "true"
-```
-
-Note that caching the registries may actually slow down the workflow running time on Windows runners.
-That is why caching the registries is disabled by default.
+By default, this caches the files in `~/.julia/artifacts/`,  `~/.julia/packages/` and `~/.julia/registries`.
```
omus: This is already outdated as also cache other depot subdirectories such as `scratchspaces`

README.md
```diff
@@ -22,24 +22,14 @@ jobs:
...
 
 ### Inputs
 
 - `cache-name` - Name used as part of the cache keys
 - `cache-artifacts` - Whether to cache `~/.julia/artifacts/`. Enabled by default.
 - `cache-packages` - Whether to cache `~/.julia/packages/`. Enabled by default.
-- `cache-registries` - Whether to cache `~/.julia/registries/`. Disabled by default.
+- `cache-registries` - Whether to cache `~/.julia/registries/`. Enabled by default.
```
omus: The README looks quite a bit different from this now. Maybe you need to merge in the latest `main`?

omus (approved):

IanButterworth:
Thanks. Based on that it does seem we can enable this by default.

There are cases where jobs do things like this
```
julia -e 'import Pkg; Pkg.add("JuliaFormatter"); format()`
```
i.e. use non-stdlib packages without a Project/Manifest

These cases will be affected by the registry update 24hrs cool-off, but in these cases the package are being used as tooling that is likely to be less critical to be up to date.

If that's the case the job can call `Pkg.Registry.update()` before `add`, or we could patch Pkg to avoid the cool-down period if on CI.

PR #66, opened 2023-09-06
Remove "test" from cache name

fonsp:
feel free to merge/close without comment :)


SaschaMann (approved):

SaschaMann:
Do we want to add the `-test-` version to the restore keys for a while (~1m?) to not invalidate all existing caches?

Issue #46, opened 2023-01-09
Julia's version is not taken into account


rikhuijzer:
Julia's version is currently not taken into account in the cache key. This probably has a negative effect on the usefulness of the compiled cache when there are multiple Julia versions being cached.

Is this something we can fix? Would it be safe for us to read out `julia -e 'println(VERSION)'`?

giordano:
The compile cache is already indexed by the julia version on the file system, no?

rikhuijzer:
Yes, but that could still mean that a cache by Julia version X results in a lot of X-specific files which are useless when restored on Julia version Y.

SaschaMann:
We could add `julia --version` to the cache key if that's useful. Particularly with #45 that moves the compiled cache to a separate cache anyway.

PR #71, opened 2023-11-19
Update cache every run. Add `/compiled` and `/logs`. Make key sensitive to matrix.

IanButterworth:
Alternative to https://github.com/julia-actions/cache/pull/45
Closes https://github.com/julia-actions/cache/issues/11
Fixes https://github.com/julia-actions/cache/issues/46

This keeps the depot in one cache file for the given key, which  is sensitive to ~julia version, OS and arch (for the usual matrix var names, and there's a note if not).~ all matrix vars and the runner.OS.
It also re-saves the cache every run, removing the previous cache that was used.
By including `/logs` it also enables automatic `Pkg.gc()` to keep the depot size down, given that the depot will be restored>used>saved>restored>used>saved.. sequentially through the runs.

Regarding the split strategy in #45 , while there are some elements of the depot that are cross-OS compatible (`/packages`, `/registries`), for precompilation to remain hot, and for automatic `Pkg.gc()` to work properly, I think the depot should be cached in lock-step as a whole.

To implement this, because neither Github or `actions/cache` expose a way to update a cache file at a key, we have to run our own deletion just before the `actions/cache` save post step. So for a given key, there will only ever be one cache retained. This reduces the chance of the automatic eviction github does evicting a useful cache over an less useful one.

## Demo

You can see it in action here https://github.com/JuliaIO/PNGFiles.jl/pull/72
Tests have been disabled, so it's just installation and loading.

First commit was with no cache files in the repo.
Second empty commit benefitted from the caching.

On linux 1.11 goes from 2 minutes to 40s

Cache sizes are all around 30 MB.

And there will only ever be the same number as the permutations of the matrix, so 15 in that repo, which is a common number.

I will note that it seems there's been a sizeable regression in precompile duration in master recently. Needs investigating.

## How to release

It'd be nice to get this widely adopted, so I'm intrigued about releasing this as a minor change.
There's no breakage to the API, except for changes to defaults.


handle_caches.jl
```diff
@@ -0,0 +1,30 @@
...
+        filter!(contains(restore_key), hits)
+        isempty(hits) && break
+        for c in hits
+            try
+                run(`gh cache delete $(split(c)[1])`)
+            catch e
+                @error e
+            end
+        end
+        append!(caches, hits)
```
rikhuijzer, line 21:
```suggestion
        append!(caches, hits)
        search_again = length(hits) == 100
```

handle_caches.jl
```diff
@@ -0,0 +1,30 @@
...
+restore_key = get(ARGS, 2, "")
+
+if func == "list"
+    println("Listing existing caches")
+    run(`gh cache list --limit 100`)
+elseif func == "rm"
+    caches = String[]
+    for _ in 1:5 # limit to avoid accidental rate limiting
+        hits = split(strip(read(`gh cache list --limit 100`, String)), keepempty=false)
+        search_again = length(hits) == 100
```
rikhuijzer, line 11:
```suggestion
```
Variable can be moved down

handle_caches.jl
```diff
@@ -0,0 +1,30 @@
...
+        append!(caches, hits)
+        search_again || break
+    end
+    if isempty(caches)
+        println("No existing caches found for restore key `$restore_key`")
+    else
+        println("$(length(caches)) existing caches deleted that match restore key `$restore_key`:")
+        println.(caches)
+    end
+end
```
rikhuijzer, line 30:
```suggestion
else
    throw(ArgumentError("Unexpected first argument: $func"))
end
```

rikhuijzer (approved):
Nice. Makes sense to me. The cache deletion logic looks a bit like a "hack", but it makes sense to me. For Javascript libraries, not updating the cache should be fine for small dependency changes. However, for Julia small dependency changes can outdate almost all compiled artifacts, so that's probably why forcing a new cache each time makes sense here?

One question regarding the cache delete logic. Isn't the most recent cache also deleted?

handle_caches.jl
IanButterworth: This isn't correct. The logic is that if the unfilitered original list is equal to 100, then there are more. After the filter we don't know

IanButterworth:
> isn't the most recent cache deleted?

No, the cache deletion step happens just before the cache/save step.

The way post jobs are queued is a little confusing.

handle_caches.jl
```diff
@@ -0,0 +1,30 @@
...
+    println("Listing existing caches")
+    run(`gh cache list --limit 100`)
+elseif func == "rm"
+    caches = String[]
+    for _ in 1:5 # limit to avoid accidental rate limiting
+        hits = split(strip(read(`gh cache list --limit 100`, String)), keepempty=false)
+        search_again = length(hits) == 100
+        filter!(contains(restore_key), hits)
+        isempty(hits) && break
+        for c in hits
```
rikhuijzer, line 14:
```suggestion
        # We can delete everything that matches the restore key because the new cache is saved later.
        for c in hits
```

IanButterworth:
Ideally we'd only save if there were changes in the depot, but in practice that will be true every run because of the usage logs being updated. And usage logs allow us to do auto Pkg.gc

It kind of puts the decision on how to manage the depot in Julia's hands, which I think makes sense.

IanButterworth:
Reminder to myself. We might want to add an input to turn off cache deletion.

README.md
```diff
@@ -34,25 +34,59 @@ To also cache `~/.julia/registries/`, use
...
+
+### Cache Retention
+
+This action automatically deletes old caches that match the first 4 fields of the above key:
+- The `cache-name` input
+- An assumed `matrix.julia-version` variable (ignored if not found)
+- The `runner.os`
+- An assumed `matrix.arch` variable (ignored if not found)
+
+Which means your caches files will not grow needlessly. Github also deletes cache files after 7 days.
```
IanButterworth, line 83:
```suggestion
Which means your caches files will not grow needlessly. Github also deletes cache files after [90 days and up to 400 days for private repos](https://docs.github.com/en/organizations/managing-organization-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-organization)
```

IanButterworth:
If you're able to review @SaschaMann let me know, otherwise I'll merge and release this as v1.4.0 tomorrow. (or let me know if you want more time)

action.yml
```diff
@@ -43,18 +49,41 @@ runs:
...
+          ${{ steps.paths.outputs.registries-path }}
+          ${{ steps.paths.outputs.scratchspaces-path }}
+          ${{ steps.paths.outputs.compiled-path }}
+          ${{ steps.paths.outputs.logs-path }}
+        # the key should be unique to the run, otherwise actions/cache will refuse to save it
+        key: ${{ inputs.cache-name }}-${{ matrix.julia-version }}-${{ runner.os }}-${{ matrix.arch }}-${{ github.run_id }}-${{ github.run_attempt }}
+        # Note: if changed, make sure the restore key is updated in the delete step too
         restore-keys: |
-          ${{ runner.os }}-${{ inputs.cache-name }}-
+          ${{ inputs.cache-name }}-${{ matrix.julia-version }}-${{ runner.os }}-${{ matrix.arch }}
```
IanButterworth, line 73: Actually `version` might be more common than `julia-version`.

Or we just be sensitive to both like this (given non-existant matrix var is interpolated as an empty string)
```suggestion
          ${{ inputs.cache-name }}-${{ matrix.version }}${{ matrix.julia-version }}-${{ runner.os }}-${{ matrix.arch }}
```

IanButterworth:
I've done some tidying. @rikhuijzer would you mind taking one last look before I merge? This is in the context of it being a non-breaking release

README.md
```diff
@@ -22,37 +22,72 @@ jobs:
...
 
-## How it works
+## How It Works
 
 This action is a wrapper around <https://github.com/actions/cache>.
 In summary, this action stores the files in the aforementioned paths in one compressed file when running for the first time.
-This cached file is then restored upon the second run.
-The benefit of this is that downloading one big file is quicker than downloading many different files from many different locations.
+This cached file is then restored upon the second run, and afterwards resaved under a new key, and the previous cache deleted.
+The benefit of this is that downloading one big file is quicker than downloading many different files from many different locations
```
rikhuijzer, line 55:
```suggestion
The benefit of caching is that downloading one big file is quicker than downloading many different files from many different locations
```
Otherwise it looks like "this" refers to the resaving under new key and previous cache deletion logic.

rikhuijzer (approved):
I have to admit that I don't fully understand everything because I find caching logic complex. Having said that, I'm in favor of going ahead and merging 👍

action.yml
```diff
@@ -43,18 +49,42 @@ runs:
...
+          ${{ steps.vars.outputs.compiled-path }}
+          ${{ steps.vars.outputs.logs-path }}
+
+        key: ${{ steps.vars.outputs.key }}
+        restore-keys: ${{ steps.vars.outputs.restore-key }}
+        enableCrossOsArchive: false
+
+    # github and actions/cache doesn't provide a way to update a cache at a given key, so we delete any
+    # that match the restore key just before saving the new cache
+    - uses: pyTooling/Actions/with-post-step@v0.4.6
```
SaschaMann, line 79: This action appears to be quite simple. Given we expect the cache action to be used across such a wide range of repos and perhaps also in contexts with higher permissions than `read`, we should consider replicating its logic in here.

At the very least, it should be pinned at a commit hash like the other actions. [This is what the request-changes status is for.]

Depending on third party actions adds a lot of attack surface to this action and thus to the ecosystem, so it should be carefully considered.

SaschaMann (changes requested):
Looks good for the most part, though I didn't have the time to review it thoroughly.

action.yml
IanButterworth: Yeah. For now I've changed it to the commit hash, but I may bring it over as others have done the same for this action, with a note for the license

IanButterworth:
Ok. I've found a better way than guessing matrix var names.. now this collates all matrix vars in the order they are entered automatically into a string.

So 
```
      matrix:
        julia-version: ['1', '^1.10.0-rc1', 'nightly']
        os: [ubuntu-latest, windows-latest, macOS-latest]
        arch: [x64, x86]
```
becomes, for instance `1-ubuntu-latest-x64`
And I included `runner.OS` just in case it's not part of the matrix.

So a full default key looks like `julia-cache-1-ubuntu-latest-x64-Linux-6961372938-1`


You can see the latest in action here
https://github.com/JuliaIO/PNGFiles.jl/actions/runs/6961372938/attempts/1?pr=72
and the re-run with a hot cache (although some of the nightlies updated)
https://github.com/JuliaIO/PNGFiles.jl/actions/runs/6961372938/attempts/2?pr=72

action.yml
rikhuijzer: Maybe something like this could work also (https://docs.github.com/en/actions/learn-github-actions/expressions#always)?

```
    - run: |
        echo "This is the post step"
        # Add your post step logic here
      shell: bash
      if: always() 
```

IanButterworth: I'm not sure how that would work for our needs. What we need to do is inject a step after the users last step, but before the cache/save step. IIUC the only way to do that is to use the `post` approach, and setting that up after the `cache` step here means we get that ordering.

rikhuijzer:
For future reference, here are the numbers from the two runs:

<details>

| Test Description | First Run | Second Run |
|------------------|----------------|----------------|
| test (nightly, macOS-latest) | 2m 56s | 3m 38s |
| test (nightly, ubuntu-latest) | 2m 1s | 38s |
| test (nightly, ubuntu-latest, ...) | 2m 0s | 1m 54s |
| test (nightly, windows-latest, ...) | 3m 5s | 1m 36s |
| test (nightly, windows-latest, ...) | 3m 4s | 1m 29s |
| test (^1.10.0-rc1, macOS-latest, ...) | 1m 59s | 2m 0s |
| test (^1.10.0-rc1, ubuntu-latest, ...) | 1m 38s | 1m 36s |
| test (^1.10.0-rc1, ubuntu-latest, ...) | 1m 35s | 1m 28s |
| test (^1.10.0-rc1, windows-latest, ...) | 2m 39s | 1m 19s |
| test (^1.10.0-rc1, windows-latest, ...) | 2m 35s | 1m 8s |
| test (1, macOS-latest, ...) | 3m 22s | 1m 20s |
| test (1, ubuntu-latest, ...) | 1m 35s | 37s |
| test (1, ubuntu-latest, ...) | 1m 32s | 40s |
| test (1, windows-latest, ...) | 2m 39s | 1m 2s |
| test (1, windows-latest, ...) | 2m 51s | 1m 21s |

(Credits to ChatGPT. I just threw in two screenshots asked for a table. I manually verified all the numbers. They are correct.)
</details>

Nice speedup  🚀

IanButterworth:
Note that some of the nightly versions changed between runs, so had to do re-precompilation.. That will be an issue until 1.11 is released.

But yeah, it's surprisingly effective on 1.9, but unfortunately that gain is lost in 1.10.0-rc1 because of more stdlibs outside of the sysimage.
I'm going to look into whether we can fix that regression by moving away from mtime sensitivity when it comes to stdlibs. I.e. reviving https://github.com/JuliaLang/julia/pull/50919

action.yml
IanButterworth: Now vendored ourself (as people have been doing elsewhere https://github.com/actions/runner/issues/1478#issuecomment-1586249440)

IanButterworth: Ugh.. vendoring it ourself is complicated.. relative paths work wrt the working directory, so CI passes here but fails when actually run on other projects. And we can't do `${{ github.action_path }}/with-post-step` apparently.

Anyone know of a way around this? 
Note that we cannot simply bring the logic into our `.yml` because it requires it to be a node action. `post` steps aren't possible on `composite` actions.

IanButterworth: Ok this works, and update the commit once squashed onto main
```
    - uses: julia-actions/cache/with-post-step@5d5ff4e210fbccb3e875c042dd7c8e6f9c31e1f0
```

IanButterworth:
Ok. Should be good to go now. Final run check (ignore that these are opposite order.. I deleted the caches after attempt #2)

| : cold cache : | : hot cache : |
| ---- | ---- | 
| https://github.com/JuliaIO/PNGFiles.jl/actions/runs/6961978935/attempts/3?pr=72 | https://github.com/JuliaIO/PNGFiles.jl/actions/runs/6961978935/attempts/2?pr=72 |
|  Screenshot 2023-11-22 at 2 56 43 PM | Screenshot 2023-11-22 at 2 53 02 PM |


Note: I noticed that the two nightly macOS runs are on the same julia commit.. so I think there might be an issue preventing macOS from avoiding re-precompilation. That can be investigated independently.

IanButterworth:
I think we may want to also include the job name in the key automatically?

IanButterworth:
Also from testing this with a self hosted runner. We can't assume `jq` is installed

IanButterworth:
Also, if there's no matrix
```
jq: error (at <stdin>:1): string ("") has no keys
```

action.yml
SaschaMann: I'm not sure vendoring comes with any benefits over using a commit hash unless that repo is deleted or something similar. I was thinking that it looked simple enough that we might be able to inline it (that usually also makes it clearer what's going on in case something goes wrong) but if that's not the case, I think pinning it to a  hash in their repo is sufficient.

IanButterworth: ok

SaschaMann:
> Also from testing this with a self hosted runner. We can't assume `jq` is installed

I'm not too worried about that. Requiring `jq` as a dependency seems fine, it's an extremely common tool and it's available on all hosted runners anyway.

IanButterworth:
Note to self. Hitting this on a private repo during the post step, which doesn't currently make sense to me..
```
Setting working directory to /home/runner/work/Foo.jl/Foo.jl
failed to run git: fatal: not a git repository (or any of the parent directories): .git

ERROR: LoadError: failed process: Process(`gh cache list --limit 100`, ProcessExited(1)) [1]
```

IanButterworth:
I'm noticing some things testing this with a larger less standard package setup, so I'm investigating.

SaschaMann (approved):
My concern above has been fixed, so from my side once you consider this good to go, I'm happy to have it merged.

IanButterworth:
Ok, some tweaks:
- Added `include-matrix` input bool to allow the user to disable automatically including the matrix. While including the matrix is a good default for standard julia package CI, a use case I have for it is where a single runner runner runs, then a matrix of many different permutations, and I want the matrix to use the cache from the first runner. So in that case ignoring the matrix is beneficial
- Pass the run repo explicitly to the `gh cache` calls as it cannot be assumed that the working directory is the repo of the run
- Using an underscore after the restore key helps to indicate the end of the part that restore keys should match. Otherwise too many caches were deleted
- Explicitly run `Pkg.gc` before save to help keep the cache size down
- List the depot tree size after restore, for info

IanButterworth:
ugh.. turns out you can just `${{ join(matrix.*, '-') }}` to get all matrix values joined with `-`.

So that avoids the need for `jq` 🎉 

I'll go ahead and merge, and release tomorrow as 1.4.0 unless I hear otherwise.

PR #74, opened 2023-11-21
add option to cache custom depot dir

IanButterworth:
Closes https://github.com/julia-actions/cache/issues/72

For now this is just a single directory entry, but if people need it it could be extended by inclusion of a delimiter.

PR #83, opened 2024-01-02
Add option for user specified Julia depot

omus:


omus:
Alternative to #74

IanButterworth:
I don't think this does the same thing as #74. That PR allows inclusion of a custom depot subdir.

IanButterworth:
It'd be good to have this field-tested before merge, if you happen to have such a setup

omus:
I've tested this against a private with multiple steps using separate Julia depots and things work as expected. I've decided to rename the input from `user-depot` to just `depot` as the Julia depot reference by this action may not be the user depot. 

One issue I've identified is that if you do use this action multiple times the default cache key will be the same for both of them as we don't include the depot input in the cache key at the moment. I'll add that in here yet but I've also identified a separate issue with the cache key which I'll file an issue for.

omus:
PR is ready for review

README.md
```diff
@@ -41,12 +41,13 @@ However note that caching the registries may mean that the registry will not be
 
 - `cache-name` - The cache key prefix. Defaults to `julia-cache-${{ github.workflow }}-${{ github.job }}`. The key body automatically includes matrix vars and the OS. Include any other parameters/details in this prefix to ensure one unique cache key per concurrent job type.
 - `include-matrix` - Whether to include the matrix values when constructing the cache key. Defaults to `true`.
-- `cache-artifacts` - Whether to cache `~/.julia/artifacts/`. Defaults to `true`.
-- `cache-packages` - Whether to cache `~/.julia/packages/`. Defaults to `true`.
-- `cache-registries` - Whether to cache `~/.julia/registries/`. Defaults to `false`. Disabled to ensure CI gets latest versions.
-- `cache-compiled` - Whether to cache `~/.julia/compiled/`. Defaults to `true`.
-- `cache-scratchspaces` - Whether to cache `~/.julia/scratchspaces/`. Defaults to `true`.
-- `cache-logs` - Whether to cache `~/.julia/logs/`. Defaults to `true`. Helps auto-`Pkg.gc()` keep the cache small.
+- `depot` - Path to the Julia [depot](https://pkgdocs.julialang.org/v1/glossary/) where cached data will be saved to and restored from. Defaults to `~/.julia`.
```
IanButterworth, line 44:
```suggestion
- `depot-path` - Path to the Julia [depot](https://pkgdocs.julialang.org/v1/glossary/) where cached data will be saved to and restored from. Defaults to `~/.julia`.
```

action.yml
```diff
@@ -91,7 +96,8 @@ runs:
         enableCrossOsArchive: false
 
     - name: list restored depot directory sizes
-      run: du -shc ~/.julia/* || true
+      if: ${{ steps.cache.outputs.cache-hit == 'true' }}
```
IanButterworth, line 99: 👍

README.md
omus: Bike shedding time: I considered do this but decided against this as we have [`JULIA_DEPOT_PATH`](https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH) which works in a similar way to the shell [`PATH`](https://en.wikipedia.org/wiki/PATH_(variable)) in that it's a list of directory names separate by a delimiter. I would be okay with calling this `depot-dir` or `depot-directory` (a little too long IMO) but I thought `depot` was clear enough.

We should probably do this however:
```suggestion
- `depot` - Path to a Julia [depot](https://pkgdocs.julialang.org/v1/glossary/) directory where cached data will be saved to and restored from. Defaults to `~/.julia`.
```

IanButterworth: I don't see the issue with including "path" in the name. It's the word used in the description?
Perhaps `depot-root`? `depot-dir` to me sounds like a directory in the depot.

omus: > I don't see the issue with including "path" in the name.

I'm concerned that someone could do something like this (which isn't supported when multiple depots are specified):
```yaml
depot-path: ${{ env.JULIA_DEPOT_PATH }}
```
In Julia the environmental variable is parsed to be `Base.DEPOT_PATH` which why I think it could be problematic.

> Perhaps `depot-root`?

I'm okay with this but I do prefer `depot` as that is a 1-1 match with the [Pkg glossary](https://pkgdocs.julialang.org/v1/glossary/):

> **Depot**: a directory on a system where various package-related resources live, including:
>
> `environments`: shared named environments (e.g. `v1.0`, `devtools`)
> `clones`: bare clones of package repositories
> `compiled`: cached compiled package images (`.ji` files)
> `config`: global configuration files (e.g. `startup.jl`)
> `dev`: default directory for package development
> `logs`: log files (e.g. `manifest_usage.toml`, `repl_history.jl`)
> `packages`: installed package versions
> `registries`: clones of registries (e.g. `General`)

omus:
I'll also look into using `JULIA_DEPOT_PATH` for the default value.

README.md
IanButterworth: Ok `depot` sounds good

omus:
Validated the latest change against a private GitHub actions workflow.

IanButterworth:
Given tests here don't test the default depot path, did your CI tests check that? Just want to make sure that chars like `~` and `/` are ok in artifact names

IanButterworth:
I checked and it worked with default settings on another repo

IanButterworth (approved):

PR #88, opened 2024-01-03
Unique cache-key for job matrix objects

### LINKED ISSUE #87: Non-unique cache key when using objects in job matrix
A GitHub action job matrix supports defining YAML objects as part of the matrix specification:
```yaml
strategy:
  matrix:
    pkg:
      - name: A
        dir: A.jl
      - name: B
        dir: B.jl
    version:
      - "1"
    runs-on:
      - ubuntu-latest
    arch:
      - x64
```
When utilizing this pattern with `julia-actions/cache` the resulting default generated cache key is `julia-cache-workflow-job-Object-1-ubuntu-latest-x64-Linux-...` when running for package A and package B. This is problematic as these keys are not unique which means we're only storing the cached depot information for the last run job.

omus:
Changes how the default cache key is generated such that YAML objects in the matrix are expanded upon and not just returned as "Object". Due to use combining so many values from the matrix I've also decided to include the key name in the cache key.

The change here should be non-breaking but will result in cache misses when users first update to this version.

Fixes #87.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -49,7 +49,7 @@ runs:
     - name: Install jq
       # Skip installation on GitHub-hosted runners:
       # https://github.com/orgs/community/discussions/48359#discussioncomment-5323864
-      if: ${{ !startsWith(runner.name, "GitHub Actions") }}
+      if: ${{ !startsWith(runner.name, 'GitHub Actions') }}
       uses: dcarbone/install-jq-action@v2.1.0
       with:
         force: false  # Skip install when an existing `jq` is present
--- a/action.yml
+++ b/action.yml
@@ -65,6 +65,7 @@ runs:
       # `matrix_key` joins all of matrix keys/values (including nested objects) to ensure that concurrent runs each use a unique cache key.
     - id: keys
       run: |
+        echo "MATRIX_JSON=$MATRIX_JSON"
         if [ "${{ inputs.include-matrix }}" == "true" ]; then
           matrix_key=$(echo "$MATRIX_JSON" | jq 'paths(type != "object") as $p | ($p | join("-")) + "=" + (getpath($p) | tostring)' | jq -rs 'join(";") | ";" + .')
         fi
--- a/action.yml
+++ b/action.yml
@@ -9,7 +9,7 @@ branding:
 inputs:
   cache-name:
     description: >-
-      The cache key prefix. The key body automatically includes the OS and, unless disabled, the matrix vars
+      The cache key prefix. The key body automatically includes the OS and, unless disabled, the matrix vars.
       Include any other parameters/details in this prefix to ensure one unique cache key per concurrent job type.
     default: julia-cache;workflow=${{ github.workflow }};job=${{ github.job }}
   include-matrix:
--- a/action.yml
+++ b/action.yml
@@ -62,18 +62,19 @@ runs:
         echo "logs-path=$L_PATH" >> $GITHUB_OUTPUT
       shell: bash
 
-      # MATRIX_STRING is a join of all matrix variables that helps concurrent runs have a unique cache key.
-      # The underscore at the end of the restore key demarks the end of the restore section. Without this
-      # a runner without a matrix has a restore key that will cause impropper clearing of caches from those
-      # with a matrix.
+      # `matrix_key` joins all of matrix keys/values (including nested objects) to ensure that concurrent runs each use a unique cache key.
     - id: keys
       run: |
-        [ "${{ inputs.include-matrix }}" == "true" ] && MATRIX_STRING="${{ join(matrix.*, '-') }}"
-        [ -n "$MATRIX_STRING" ] && MATRIX_STRING="-${MATRIX_STRING}"
-        RESTORE_KEY="${{ inputs.cache-name }}-${{ runner.os }}${MATRIX_STRING}_"
-        echo "restore-key=${RESTORE_KEY}" >> $GITHUB_OUTPUT
-        echo "key=${RESTORE_KEY}${{ github.run_id }}-${{ github.run_attempt }}" >> $GITHUB_OUTPUT
+        if [ "${{ inputs.include-matrix }}" == "true" ]; then
+          matrix_key=$(echo "$MATRIX_JSON" | jq 'paths(type != "object") as $p | ($p | join("-")) + "=" + (getpath($p) | tostring)' | jq -rs 'join(";") | ";" + .)
+        fi
+        restore_key="${{ inputs.cache-name }};workflow=${{ github.workflow }};job=${{ github.job }};os=${{ runner.os }}${matrix_key};"
+        key="${restore_key}run_id=${{ github.run_id }};run_attempt=${{ github.run_attempt }}"
+        echo "restore-key=${restore_key}" >> $GITHUB_OUTPUT
+        echo "key=${key}" >> $GITHUB_OUTPUT
       shell: bash
+      env:
+        MATRIX_JSON: ${{ toJSON(matrix) }}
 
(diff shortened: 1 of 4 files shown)


omus:
Still have some details to work out with the tests. Would be nice to have the tests not require `cache-name`.

IanButterworth:
We can't really use `jq` as it's not guaranteed to be installed on self-hosted runners

IanButterworth:
If we did I think we'd have to make a breaking change, which I wanted to avoid so that the ecosystem gets the benefit without changing anything

action.yml
```diff
@@ -62,18 +70,21 @@ runs:
...
-        [ -n "$MATRIX_STRING" ] && MATRIX_STRING="-${MATRIX_STRING}"
-        RESTORE_KEY="${{ inputs.cache-name }}-${{ runner.os }}${MATRIX_STRING}_"
-        echo "restore-key=${RESTORE_KEY}" >> $GITHUB_OUTPUT
-        echo "key=${RESTORE_KEY}${{ github.run_id }}-${{ github.run_attempt }}" >> $GITHUB_OUTPUT
+        # `matrix_key` joins all of matrix keys/values (including nested objects) to ensure that concurrent runs each use a unique cache key.
+        # When `matrix` isn't set for the job then `MATRIX_JSON=null`.
+        if [ "${{ inputs.include-matrix }}" == "true" ] && [ "$MATRIX_JSON" != "null" ]; then
+          matrix_key=$(echo "$MATRIX_JSON" | jq 'paths(type != "object") as $p | ($p | join("-")) + "=" + (getpath($p) | tostring)' | jq -rs 'join(";") | . + ";"')
+        fi
+        restore_key="${{ inputs.cache-name }};os=${{ runner.os }};${matrix_key}"
```
omus, line 81: I think we can drop `runner.os` if a user wants multiple OS support they'd either add the OS into the `matrix` or they would need to use a different workflow name or job name:
```suggestion
        restore_key="${{ inputs.cache-name }};${matrix_key}"
```

IanButterworth: Force including `runner.os` is mostly a failsafe. I generally agree but I worried about some edge case configuration

IanButterworth: I'm not sure I follow this point
> workflow name and job name to no longer be specified in the cache-name

WIthout it as a default, if two workflows/jobs have the same matrix then they'd share the same cache key

omus: I'm advocating for the workflow/job to not be a part of the `cache-name` default but rather be a mandatory part of this step. I originally did that in this PR but discovered that the tests for this action require require these values to be overridden so that we can have separate save/restore jobs.

IanButterworth: Ok, but forcing them into the cache key means users can't use caches across different workflow/jobs if they wanted to (without having to manually override the full cache-key as you propose in https://github.com/julia-actions/cache/issues/89)

omus: Very true. I think having the option of supplying a full `cache-key` and also returning the generated cache key as an output may allow the use of caches across different jobs and still making these portions be mandatory. I appreciate you bringing up these alternate use cases so I have a clearer understanding on some implicit design of this action.

IanButterworth: Yeah, though I didn't consolidate it down to a clear list, I did try to think through a user specification here, the key points that come to mind being:
- not breaking existing usages
- being sensibly overridable for different approaches
- working for standard julia CI approaches
- being fail-safe for `/compiled` cache issues wrt. platform

Also bear in mind that the cache-key has the workflow/job in it, but the lookup key doesn't. That means it's possible to restore without paying attention to workflow/job, but will always be able to save. (it'd be nice if `actions/cache` can force save, but it can't)

omus:
PR's ready for review. I'll mention that introducing the matrix keys into the cache key isn't strictly necessary but I think it's helpful when you have matrices with lots of options. It does make the cache names in GitHub actions much longer but I think the additional context which is added is more helpful for humans in identifying which cache entry they want to manually delete.

action.yml
```diff
@@ -62,18 +70,21 @@ runs:
...
       run: |
-        [ "${{ inputs.include-matrix }}" == "true" ] && MATRIX_STRING="${{ join(matrix.*, '-') }}"
-        [ -n "$MATRIX_STRING" ] && MATRIX_STRING="-${MATRIX_STRING}"
-        RESTORE_KEY="${{ inputs.cache-name }}-${{ runner.os }}${MATRIX_STRING}_"
-        echo "restore-key=${RESTORE_KEY}" >> $GITHUB_OUTPUT
-        echo "key=${RESTORE_KEY}${{ github.run_id }}-${{ github.run_attempt }}" >> $GITHUB_OUTPUT
+        # `matrix_key` joins all of matrix keys/values (including nested objects) to ensure that concurrent runs each use a unique cache key.
+        # When `matrix` isn't set for the job then `MATRIX_JSON=null`.
+        if [ "${{ inputs.include-matrix }}" == "true" ] && [ "$MATRIX_JSON" != "null" ]; then
+          matrix_key=$(echo "$MATRIX_JSON" | jq 'paths(type != "object") as $p | ($p | join("-")) + "=" + (getpath($p) | tostring)' | jq -rs 'join(";") | . + ";"')
```
omus, line 79: There's lots to unpack here but ultimately it takes JSON key/values and joins the keys with all of it's ancestor keys and then joins them into a single string (e.g. `{"pkg": {"name": "A", "dir": "A.jl"}}` to `pkg-name=A;pkg-dir=A.jl;`). One subtle thing here is that the jq string concatenation only works with strings so entries like:
```yaml
threads:
- 1
- 2
```
Require the use of `tostring`.

action.yml
```diff
@@ -9,7 +9,7 @@ branding:
 inputs:
   cache-name:
     description: 'The cache key prefix. Unless disabled the key body automatically includes matrix vars, and the OS. Include any other parameters/details in this prefix to ensure one unique cache key per concurrent job type.'
```
IanButterworth, line 11: Might as well fix this detail here
```suggestion
    description: 'The cache key prefix. The key body automatically includes the OS and, unless disabled, the matrix vars. Include any other parameters/details in this prefix to ensure one unique cache key per concurrent job type.'
```

omus:
Noticed a README update I need to do

action.yml
```diff
@@ -9,9 +9,9 @@ branding:
 inputs:
   cache-name:
     description: >-
-      The cache key prefix. Unless disabled the key body automatically includes matrix vars, and the OS.
+      The cache key prefix. The key body automatically includes the OS and, unless disabled, the matrix vars
```
IanButterworth, line 12:
```suggestion
      The cache key prefix. The key body automatically includes the OS and, unless disabled, the matrix vars.
```

IanButterworth (approved):

musoke:
FYI: this change seems to have broken one of my CI runs.  I think the problem is that I'm running in a container on a GitHub hosted runner, so the jq isn't installed.  

here's a log: https://github.com/musoke/WolframExpr.jl/actions/runs/7532058237/job/20501935392#step:5:65

action.yml
```diff
@@ -51,6 +51,14 @@ outputs:
 runs:
   using: 'composite'
   steps:
+    - name: Install jq
+      # Skip installation on GitHub-hosted runners:
+      # https://github.com/orgs/community/discussions/48359#discussioncomment-5323864
+      if: ${{ !startsWith(runner.name, 'GitHub Actions') }}
```
IanButterworth, line 57: I guess this can't be relied on so maybe we just remove this line @omus ?

musoke: Wouldn't it be simplest to directly check if jq is installed?  Maybe there's another subtlety I am missing.

omus: Tracking issue: https://github.com/julia-actions/cache/issues/104

PR #91, opened 2024-01-06
make depot if not restored

IanButterworth:
This helps other following actions know that this action ran
i.e. https://github.com/julia-actions/julia-buildpkg/pull/41

Alternatively we could set an env var, but this seems logical to me

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -117,9 +117,12 @@ runs:
         restore-keys: ${{ steps.keys.outputs.restore-key }}
         enableCrossOsArchive: false
 
-    - name: list restored depot directory sizes
-      if: ${{ steps.cache.outputs.cache-hit == 'true' }}
-      run: du -shc ${{ steps.paths.outputs.depot }}/* || true
+    # if it wasn't restored make the depot anyway as a signal that this action ran
+    # for other julia actions to check, like https://github.com/julia-actions/julia-buildpkg/pull/41
+    - name: make depot if not restored, then list depot directory sizes
+      run: |
+        mkdir -p ${{ steps.paths.outputs.depot }}
+        du -shc ${{ steps.paths.outputs.depot }}/* || true
       shell: bash
 
     # github and actions/cache doesn't provide a way to update a cache at a given key, so we delete any


action.yml
```diff
@@ -117,9 +117,12 @@ runs:
...
         enableCrossOsArchive: false
 
-    - name: list restored depot directory sizes
-      if: ${{ steps.cache.outputs.cache-hit == 'true' }}
-      run: du -shc ${{ steps.paths.outputs.depot }}/* || true
+    # if it wasn't restored make the depot anyway as a signal that this action ran
+    # for other julia actions to check, like https://github.com/julia-actions/julia-buildpkg/pull/41
+    - name: make depot if not restored, then list depot directory sizes
+      run: |
+        mkdir -p ${{ steps.paths.outputs.depot }}
```
omus, line 124: I'd personally make this a separate step which only occurs on a cache miss

IanButterworth: So it seems that `steps.cache.outputs.cache-hit` is only `true` [if the full key is hit](https://github.com/actions/cache#outputs). We only ever hit partial hits by design, so will always be `false`, so the only way to do this would be to check if the dir exists in the cmd rather than do an `if`. 

The current setup is robust to this, so maybe we just leave this PR as is?

omus: You are correct. This would be possible to determine if `actions/cache` had the same [outputs as `actions/cache/restore`](https://github.com/actions/cache/blob/main/restore/README.md#outputs). Oh well

omus (approved):

action.yml
```diff
@@ -120,9 +120,12 @@ runs:
...
+        du -shc ${{ steps.paths.outputs.depot }}/* || true
```
fredrikekre, line 128: Perhaps this should be guarded behind something, otherwise it looks a bit like an error in the logs when the folder is empty:
```
du: cannot access '/home/runner/.julia/*': No such file or directory
```

IanButterworth: Yeah I tried to figure that out but couldn't. It's the contents that doesn't exist not the dir.

PR #93, opened 2024-01-09
URL encode any invalid key characters

omus:
Noticed in a private workflow that when a matrix values contain commas the cache restore step fails with:
```
Error: Key Validation Error: <redacted-key> cannot contain commas.
```
I ended up [tracking down where the key restrictions are implemented](https://github.com/actions/toolkit/blob/5430c5d84832076372990c7c27f900878ff66dc9/packages/cache/src/cache.ts#L32-L44) and found that there are only two restrictions:
- Limited to a length of 512 characters
- Cannot use commas

I've opted to URL encode the comma character only to keep the artifact key as human-readable as possible while still indicating the presence of a comma. I'm not as worried about the key length limit as my artifact key was only 286 characters and was using a pretty complicated job matrix with included paths.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -91,6 +91,7 @@ jobs:
         dep:
           - name: pandoc_jll
             version: "3"
+            invalid-chars: ","  # Use invalid characters in job matrix to ensure we escape them
         os:
           - ubuntu-latest
           - windows-latest
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -39,6 +39,7 @@ jobs:
         dep:
           - name: pandoc_jll
             version: "3"
+            invalid-chars: ","  # Use invalid characters in job matrix to ensure we escape them
         os:
           - ubuntu-latest
           - windows-latest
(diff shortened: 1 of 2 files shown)


IanButterworth:
LGTM but worth a test? Can we just put a comma in the CI matrix somewhere?

omus:
> LGTM but worth a test? Can we just put a comma in the CI matrix somewhere?

Will do. Would love to get some unit tests setup to make it more obvious what we're all testing for

IanButterworth (approved):

omus:
The Windows CI job failed due to network issues. Will push an empty commit to try again:
```
Run using Pkg; Pkg.add(PackageSpec(name="pandoc_jll", version="3"))
  
  Installing known registries into `/tmp/julia-depot`
ERROR: LoadError: could not download https://pkg.julialang.org/registry/23338594-aafe-5451-b93e-139f81909106/65918196649b53895b4fe97bb7661f82807c9e49 
Exception: RequestError: HTTP/1.1 404 Not Found while requesting https://pkg.julialang.org/registry/23338594-aafe-5451-b93e-139f81909106/65918196649b53895b4fe97bb7661f82807c9e49
```

omus:
@IanButterworth feel free to merge. I don't have write access.

PR #96, opened 2024-01-12
Test cache cleanup with multiple workflows

omus:
Validating the work done in #95 with a branch that exists in the same repo. It appears the forks with workflow dispatch events don't work well.


omus:
I no longer believe `workflow_dispatch` will work here as I think the GHA caching don't treat those workflows as on the same branch and thereby cannot load the cache

omus:
Validated this is working

PR #97, opened 2024-01-12
Delete cache entries only on the workflow branch

omus:
I've noticed that the post-action step which automatically cleans up old cache entries accidentally cleans up entries created on other branches. This reduces performance with multiple PRs are being developed concurrently especially in the scenario where a PR deletes the cache entry for the main branch resulting in all new PRs having cache misses until a PR has been merged.

This PR updates the handle_caches.jl script to only delete caches which match the restore key using the same git ref as the executing workflow. This should result in workflows run on the main branch only deleting cache entries on main and PRs only deleting cache entries for that PR.

For more details about cache isolation in regards to branches see: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -52,7 +52,7 @@ jobs:
       id: cache
       uses: ./
       with:
-        cache-name: ${{ needs.generate-key.outputs.cache-name }}
+        cache-name: ${{ needs.generate-key.outputs.cache-name }}-matrix
         delete-old-caches: required
     - name: Check no artifacts dir
       shell: 'julia --color=yes {0}'
@@ -73,7 +73,7 @@ jobs:
       id: cache
       uses: ./
       with:
-        cache-name: ${{ needs.generate-key.outputs.cache-name }}
+        cache-name: ${{ needs.generate-key.outputs.cache-name }}-nomatrix
         delete-old-caches: required
     - name: Check no artifacts dir
       shell: 'julia --color=yes {0}'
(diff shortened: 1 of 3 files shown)


IanButterworth (approved):
Sounds great. Makes sense

.github/workflows/CI.yml
```diff
@@ -106,7 +110,9 @@ jobs:
       id: cache
       uses: ./
       with:
-        cache-name: ${{ needs.generate-key.outputs.cache-name }}
+        cache-name: ${{ needs.generate-key.outputs.cache-name }}-matrix
+        # Cannot require a successful cache delete on forked PRs as the permissions for actions is limited to read
+        delete-old-caches: ${{ github.event.pull_request.head.repo.full_name != 'julia-actions/cache' && 'false' || 'required' }}
```
omus, line 115: GitHub does allow repo administrators to allow  
fork PRs to have write permissions: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#changing-the-permissions-in-a-forked-repository. I think it's best to leave things as is but it's good to know this is an option

omus:
I'll hold off on merging until I triple check the non-fork version is working. Also, this may be a nicer way to detect a fork: https://github.com/orgs/community/discussions/25217#discussioncomment-3246904

IanButterworth:
I think this fixes #98.

Probably good to add 1.0 CI as these actions should probably work back to there, if possible.

omus:
> I think this fixes #98.
> 
> Probably good to add 1.0 CI as these actions should probably work back to there, if possible.

I'll save this for another PR as there are a few additional changes required to get this to work

Issue #104, opened 2024-01-16
Cannot assume `jq` is installed on runners named `GitHub Actions` 


IanButterworth:
Posted by @musoke here https://github.com/julia-actions/cache/pull/88#issuecomment-1892663926

----

FYI: this change seems to have broken one of my CI runs. I think the problem is that I'm running in a container on a GitHub hosted runner, so jq isn't installed.
v1.4.1 works but v1.5.1 does not.

here's a log: https://github.com/musoke/WolframExpr.jl/actions/runs/7532058237/job/20501935392#step:5:65

musoke:
Does that `jq` install action work inside containers?

Why does check for a GitHub hosted runner exist? Presumably this is because those runners already have `jq` installed.  Is there a problem with trying to install `jq` anyway?  It looks like it is set not to force installation if it is already installed.  Perhaps those checks are brittle too?

musoke:
I'm happy submit a PR if there is a clear solution.

PR #105, opened 2024-01-16
fix: Attempt to install jq on all runners

musoke:
Fixes: https://github.com/julia-actions/cache/issues/104

I took the simplest approach, of just removing the check on whether the runner is hosted by GitHub.  This seems to work for both [non-container](https://github.com/musoke/UltraDark.jl/actions/runs/7547005193/job/20546021767#step:4:1) and [container](https://github.com/musoke/WolframExpr.jl/actions/runs/7546831404/job/20545463332#step:6:34) jobs.

There may be other cases and subtleties that I am missing.

The one caveat is that the container has to have curl installed; this was easy to do but may be worth adding to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ By default all depot directories called out below are cached.
 ### Requirements
 
 This action uses [`jq`](https://github.com/jqlang/jq) to parse JSON.
-`jq` is installed by default in GitHUb-hosted runners.
+`jq` is installed by default in GitHub-hosted runners.
 [`dcarbone/install-jq-action`](https://github.com/dcarbone/install-jq-action) is used to check that `jq` is available and install it if not.
 **Note:** installing `jq` with `dcarbone/install-jq-action` requires that curl is available; this may not be the case in custom containers.
 
--- a/README.md
+++ b/README.md
@@ -29,6 +29,13 @@ jobs:
 
 By default all depot directories called out below are cached.
 
+### Requirements
+
+This action uses [`jq`](https://github.com/jqlang/jq) to parse JSON.
+`jq` is installed by default in GitHUb-hosted runners.
+[`dcarbone/install-jq-action`](https://github.com/dcarbone/install-jq-action) is used to check that `jq` is available and install it if not.
+**Note:** installing `jq` with `dcarbone/install-jq-action` requires that curl is available; this may not be the case in custom containers.
+
 ### Optional Inputs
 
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -52,9 +52,9 @@ runs:
...
     - name: Install jq
-      # Skip installation on GitHub-hosted runners:
-      # https://github.com/orgs/community/discussions/48359#discussioncomment-5323864
-      if: ${{ !startsWith(runner.name, 'GitHub Actions') }}
+      # # Skip installation on GitHub-hosted runners:
+      # # https://github.com/orgs/community/discussions/48359#discussioncomment-5323864
+      # if: ${{ !startsWith(runner.name, 'GitHub Actions') }}
       uses: dcarbone/install-jq-action@v2.1.0
       with:
         force: false  # Skip install when an existing `jq` is present
```
musoke, line 60: `force: false` means that `dcarbone/install-jq-action` does not attempt to install `jq` if it already exists. This seems to be sufficient to avoid reinstalling it on GitHub-hosted images, [which already have jq installed](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#installed-apt-packages)?

giordano: The fact that jq is required should be clearly documented though

action.yml
```diff
@@ -52,9 +52,9 @@ runs:
   using: 'composite'
   steps:
     - name: Install jq
-      # Skip installation on GitHub-hosted runners:
-      # https://github.com/orgs/community/discussions/48359#discussioncomment-5323864
-      if: ${{ !startsWith(runner.name, 'GitHub Actions') }}
+      # # Skip installation on GitHub-hosted runners:
+      # # https://github.com/orgs/community/discussions/48359#discussioncomment-5323864
+      # if: ${{ !startsWith(runner.name, 'GitHub Actions') }}
```
IanButterworth, line 57:
```suggestion
```

IanButterworth:
Please do add the curl note to the readme

IanButterworth (approved):
LGTM. @omus

giordano (approved):

omus (approved):
Thanks for this. Let's get this working first and later we can try find a way avoid the install if it's slowing things down.

README.md
```diff
@@ -29,6 +29,13 @@ jobs:
 
 By default all depot directories called out below are cached.
 
+### Requirements
+
+This action uses [`jq`](https://github.com/jqlang/jq) to parse JSON.
+`jq` is installed by default in GitHUb-hosted runners.
```
omus, line 35:
```suggestion
`jq` is installed by default in GitHub-hosted runners.
```

PR #106, opened 2024-01-18
Pin third party action to hash

SaschaMann:
Third party actions should never not be pinned to a hash. Otherwise, in case the action repo is taken over by a malicious actor, they can change what runs in all of the workflows that julia-actions/cache is used in as well. Pinning to a hash prevents that.


IanButterworth (approved):

omus (approved):
TIL

PR #108, opened 2024-01-19
Mention why the extra permissions are required

rikhuijzer:
The new permissions
```yml
permissions:
  actions: write
  contents: read
```
can be a bit of a hassle to setup for many repositories.

It's probably good to clarify why these permissions are useful in order to avoid confusion and possibly frustration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,9 @@ This cached file is then restored upon the second run, and afterwards resaved un
 The benefit of caching is that downloading one big file is quicker than downloading many different files from many different locations
 and precompiling them.
 
-By default, this actions removes the previous cache to make the cached files are as useful as possible.
-A reason for this is that, unlike many other languages, updating one dependency in Julia can outdate large amounts of precompiled code.
-By removing the old cache, this Action ensures that caches will be updated for small changes to the dependencies.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,6 +62,10 @@ This cached file is then restored upon the second run, and afterwards resaved un
 The benefit of caching is that downloading one big file is quicker than downloading many different files from many different locations
 and precompiling them.
 
+By default, this actions removes the previous cache to make the cached files are as useful as possible.
+A reason for this is that, unlike many other languages, updating one dependency in Julia can outdate large amounts of precompiled code.
+By removing the old cache, this Action ensures that caches will be updated for small changes to the dependencies.
```
IanButterworth, line 67:
```suggestion
By default, this actions removes caches that were previously made by CI jobs on the same branch with the same restore key. While GitHub automatically removes old caches when the repository cache allocation is full, it is more efficient to delete like this to avoid penalizing caching on jobs that are run less frequently.
```

rikhuijzer: I've implemented this in 6d8227a with some minor modifications to get rid of the double negative around "avoid penalizing caching ... that are run less frequently"

IanButterworth (approved):

PR #121, opened 2024-04-03
Update actions/cache to v3.3.3

### LINKED ISSUE #119: Why are we using this specific commit of `actions/cache`?
The [`action.yml`](https://github.com/julia-actions/cache/blob/main/action.yml) file in this repo currently pins `actions/cache` to the following commit:

https://github.com/julia-actions/cache/blob/0413f3b1aed10bc4d367678d38a695e81c7c09f5/action.yml#L118-L118

Now, it is very good that we are pinning all external actions to the full-length commit hash.

However, what is unusual here is that specific commit we pin to doesn't seem to correspond to any tag or branch; if you go to https://github.com/actions/cache/commit/4d4ae6ae148a43d0fd1eda1800170683e9882738 you see the following message:

> This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.

This isn't necessarily a problem, but it seems a bit unusual. @IanButterworth Would you happen to know why we are using this specific commit of `actions/cache`?

IIUC, this was introduced in https://github.com/julia-actions/cache/pull/71.

cc: @IanButterworth @omus @SaschaMann @rikhuijzer

IanButterworth:
Ideally we can merge this as a patch release before https://github.com/julia-actions/cache/pull/120

Fixes #119 

As pointed out in https://github.com/julia-actions/cache/issues/119 this was erroneously left in #71 on a commit that had dependencies updated that didn't correspond with a `actions/cache` release.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -115,7 +115,7 @@ runs:
       env:
         MATRIX_JSON: ${{ toJSON(matrix) }}
 
-    - uses: actions/cache@4d4ae6ae148a43d0fd1eda1800170683e9882738
+    - uses: actions/cache@e12d46a63a90f2fae62d114769bbf2a179198b5c # v3.3.3
       id: cache
       with:
         path: |


IanButterworth:
@omus are we happy taking green CI as enough here?

Apologies for letting that commit slip in (#119)

action.yml
```diff
@@ -115,7 +115,7 @@ runs:
       env:
         MATRIX_JSON: ${{ toJSON(matrix) }}
 
-    - uses: actions/cache@4d4ae6ae148a43d0fd1eda1800170683e9882738
+    - uses: actions/cache@e12d46a63a90f2fae62d114769bbf2a179198b5c # v3.3.3
```
DilumAluthge, line 118: Just a note for people following along. Dependabot will automatically update this human-readable version number in future Dependabot PRs.

IanButterworth:
I'll hold off a release until tomorrow incase you have further testing you'd want to do @omus

Issue #109, opened 2024-01-23
Update NodejS version from `node16` to `node20`


giordano:
Node 16 is deprecated
https://github.com/julia-actions/cache/blob/b1b004ffca7da8ece08fe42c9ee0ec4fc2cc5ca5/action.yml#L118
https://github.com/julia-actions/cache/blob/b1b004ffca7da8ece08fe42c9ee0ec4fc2cc5ca5/action.yml#L139
https://github.com/julia-actions/cache/blob/b1b004ffca7da8ece08fe42c9ee0ec4fc2cc5ca5/action.yml#L150
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

giordano:
#107 probably addresses the issue for `actions/cache`, but `pyTooling/Actions/with-post-step` seems to be stuck to node16 upstream.

DilumAluthge:
Here's the issue I opened upstream for a new release of `pyTooling/Actions` that includes the `node20` bump: https://github.com/pyTooling/Actions/issues/77

DilumAluthge:
This issue will be fixed by the combination of https://github.com/julia-actions/cache/pull/123 and https://github.com/julia-actions/cache/pull/125.

DilumAluthge:
#123 and #127 have been merged.

DilumAluthge:
This issue has been fixed by the combination of https://github.com/julia-actions/cache/pull/123, https://github.com/julia-actions/cache/pull/127, and https://github.com/julia-actions/cache/releases/tag/v2.0.0.

I've already created the `v2.0.0` tag. I'll create the `v2` tag shortly.

DilumAluthge:
I've created the `v2` tag.

PR #125, opened 2024-04-05
Switch to using `pyTooling/Actions/with-post-step@tip`

DilumAluthge:
Ref #109 

The latest numbered version tag on [`pyTooling/Actions`](https://github.com/pyTooling/Actions) is `v1.0.1`. `v1.0.1` uses `node16`: https://github.com/pyTooling/Actions/blob/60281e01e2514114849e490153aa0a3627994077/with-post-step/action.yml#L39-L40

So, this PR switches from the `v1.0.1` tag to the https://github.com/pyTooling/Actions/commit/3b95a369557b3048b071dff2349f22c8871dc9f9 commit, which is the commit to which the `tip` tag currently points. The https://github.com/pyTooling/Actions/commit/3b95a369557b3048b071dff2349f22c8871dc9f9 commit uses `node20`: https://github.com/pyTooling/Actions/blob/3b95a369557b3048b071dff2349f22c8871dc9f9/with-post-step/action.yml#L39-L40

---

This is a breaking change, because the https://github.com/pyTooling/Actions/commit/3b95a369557b3048b071dff2349f22c8871dc9f9 commit upgrades the `pyTooling/Actions/with-post-step` action from `node16` to `node20`.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -136,7 +136,8 @@ runs:
     # that match the restore key just before saving the new cache
 
     # Not windows
-    - uses: pyTooling/Actions/with-post-step@60281e01e2514114849e490153aa0a3627994077 # v1.0.1
+    # TODO: switch back from the `tip` tag to a numbered version tag `vX.Y.Z`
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Replaced by #127

PR #130, opened 2024-05-21
Fix typo in `cache-name` note

rikhuijzer:
The old text contained a typo it seems because it said: "if you do not use a matrix for concurrency you could make `cache-name` such that it is unique".


IanButterworth (changes requested):
The current note is correct. If you don't use a matrix then the user is required to ensure the name handles uniqueness across concurrent jobs.

IanButterworth (approved):

PR #131, opened 2024-05-21
Add `cache-name` to main example

rikhuijzer:
This patch suggests to add `cache-name` to the main example since most Julia projects test multiple versions in a matrix.


IanButterworth:
See https://github.com/julia-actions/cache/pull/130#pullrequestreview-2068623411

Issue #113, opened 2024-03-21
Job failures should not delete branch cache


omus:
If the CI job you are running with `julia-actions/cache` fails with `delete-old-caches: true` the cache entry used by that CI job will be deleted and *no* new cache entry will be uploaded as the `actions/cache` post-step only runs on [successful jobs](https://github.com/actions/cache/blob/0c45773b623bea8c8e75f6c82b208c3cf94ea4f9/action.yml#L40).

This results in slower subsequent re-runs as we'll end up restoring a cache entry from the base branch if one is available.

omus:
Usually the `post` step is conditionally run based upon the [`post-if`](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runspost-if) setting. As [`pyTooling/Actions/with-post-step` doesn't specify this](https://github.com/pyTooling/Actions/blob/60281e01e2514114849e490153aa0a3627994077/with-post-step/action.yml#L42) I'm not sure we can fix can this beyond setting `save-always`.

omus:
I think the way we'll have to address this is to always leave the latest cache entry remaining. That way if we are cleaning up cache entries on a failed job we avoid cleaning up the last entry.

PR #135, opened 2024-06-18
Fix issues with custom cache eviction

omus:
Fixes https://github.com/julia-actions/cache/issues/113

In https://github.com/julia-actions/cache/pull/71 when we introduced saving cache entries for each matrix job we also introduced a manual cache eviction step to attempt to evict unnecessary cache entries in a better way than the default GitHub behavior (evict the oldest). Unfortunately, we have ended up being a little overly aggressive in our cache eviction policy which has resulted in the following problems:

1. We assume that a cache entry will be saved upon every job run so it is safe to evict all cache entries on that branch before we save the new entry. For failed jobs no new cache entry is saved so we end up starting from scratch on subsequent runs. In this PR I've updated the behavior to always keep the latest cache entry (possibly in addition to the new cache entry) so failures like this can use the last saved entry.
2. We perform cache eviction on entries from the default branch. As GitHub action caches can only be pulled from the current branch or the base branch the default branch can be rather important for falling back on for a base cache entry. When new commits are added to the default branch the old cache entries are evicted which can leave PRs with no base cache to fall back on. I now avoid performing manual cache eviction on the default branch to avoid this kind of problem.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -148,13 +148,6 @@ runs:
     # - https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache
     # - https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
 
-    - name: Debug
-      shell: bash
-      run: |
-        echo "- ${{ github.event.repository.default_branch }}"
-        echo "- ${{ github.ref }}"
-
-
     # Not windows
     - uses: pyTooling/Actions/with-post-step@e9d0dc3dba9fda45f195946858708f60c0240caf # v1.0.5
       if: ${{ inputs.delete-old-caches != 'false' &&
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -132,12 +132,34 @@ runs:
...
+      run: |
+        echo "- ${{ github.event.repository.default_branch }}"
+        echo "- ${{ github.ref }}"
+
 
     # Not windows
     - uses: pyTooling/Actions/with-post-step@e9d0dc3dba9fda45f195946858708f60c0240caf # v1.0.5
-      if: ${{ inputs.delete-old-caches != 'false' && runner.OS != 'Windows' }}
+      if: ${{ inputs.delete-old-caches != 'false' &&
+              github.ref != format('refs/heads/{0}', github.event.repository.default_branch) &&
```
omus, line 161: The `default_branch` parameter is definitely present for events `push` and `pull_request`. Other events which possibly may not have this field would result in the comparison `github.ref != 'refs/heads/'` which would always evaluate to `true`

omus:
I've done some testing against this change on a private repo and have been happy with the results. We still have the issue with the scenario where a job which fails doesn't save a cache entry but since we can at least reliably fallback on the default branch cache entry things are in a much better state.

IanButterworth (approved):

IanButterworth:
Sounds good and thanks for the external testing

PR #143, opened 2024-08-31
Soft fail registry update

### LINKED ISSUE #142: Assumes Julia is installed
The registry update step assumes julia is installed and fails the job if it's not. 

cc. @giordano

IanButterworth:
Fixes #142

Would this mark the job as failed even if steps succeeded afterwards?

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -140,6 +140,7 @@ runs:
     # Pkg operations. So this is only necessary for newer julia versions.
     - name: Update any cached registries
       if: ${{ inputs.cache-registries == 'true' }}
+      continue-on-error: true
       run: |
         if [ -d "${{ steps.paths.outputs.depot }}/registries" ] && [ -n "$(ls -A "${{ steps.paths.outputs.depot }}/registries")" ]; then
           echo "Registries directory exists and is non-empty. Updating any registries"


IanButterworth:
According to ChatGPT this should work:


In GitHub Actions, if a step within a job is marked with `continue-on-error: true` and it fails, the job will continue executing the subsequent steps. Whether the job as a whole is marked as successful or not depends on the outcomes of the remaining steps.

Here's what happens:

1. **Step with `continue-on-error: true` fails**: This failure is ignored, and the job proceeds to the next steps.
   
2. **Subsequent steps succeed**: If all the subsequent steps succeed, then the overall job will be marked as successful, despite the earlier step failure. This is because the failure of a step with `continue-on-error: true` does not contribute to the final job status.

3. **Subsequent steps fail**: If any of the subsequent steps fail (and they are not marked with `continue-on-error: true`), then the job will be marked as failed.

In summary, a job can be marked as successful if a step with `continue-on-error: true` fails, as long as the steps that follow it succeed.

IanButterworth:
Successfully tested here

https://github.com/IanButterworth/PowerMonitor.jl/actions/runs/10644589690/job/29509494273#step:3:140

giordano (approved):
Looks good to me, thanks! But maybe add a comment to explain why we want to ignore errors on this step?

PR #149, opened 2024-10-31
Avoid evaluation of command substitution in input

giordano:
In the body of the bash script, `${{ inputs }}` parameters are often quoted with double quotes, which allow command substitution.  This replaces double quotes with single quotes to prevent that and avoid command substitution.

First spotted in https://github.com/EnzymeAD/Reactant.jl/actions/runs/11603689684/job/32311135543#step:3:87. Note that the `restore_key` is set to
```
  restore_key="julia-cache;workflow=Format `main`;job=format;os=Linux;${matrix_key}"
```
Note the workflow is called literally `` Format `main` ``, which inside double quotes is evaluated, leading to an error in this case, but it opens the door to arbitrary code execution just by tweaking the workflow name.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -58,8 +58,8 @@ runs:
 
     - id: paths
       run: |
-        if [ -n "${{ inputs.depot }}" ]; then
-            depot="${{ inputs.depot }}"
+        if [ -n '${{ inputs.depot }}' ]; then
+            depot='${{ inputs.depot }}'
         elif [ -n "$JULIA_DEPOT_PATH" ]; then
             # Use the first depot path
             depot=$(echo $JULIA_DEPOT_PATH | cut -d$PATH_DELIMITER -f1)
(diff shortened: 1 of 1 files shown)


omus:
An alternate solution would be to environmental variables for each input which allows the use of double quotes still.
```
env:
  depot: ${{ inputs.depot }}
```
I'm fine with the single-quote fix here but I do think we need a comment about it in `action.yaml` so we don't accidentally re-introduce double-quotes.

action.yml
```diff
@@ -190,7 +190,7 @@ runs:
               runner.OS == 'Windows' }}
       with:
         main: echo ""
-        post: cd %GITHUB_ACTION_PATH% && julia handle_caches.jl rm "${{ github.repository }}" "${{ steps.keys.outputs.restore-key }}" "${{ github.ref }}" "${{ inputs.delete-old-caches != 'required' }}"
+        post: cd %GITHUB_ACTION_PATH% && julia handle_caches.jl rm '${{ github.repository }}' '${{ steps.keys.outputs.restore-key }}' '${{ github.ref }}' "${{ inputs.delete-old-caches != 'required' }}"
```
omus, line 193: May as well use single quotes for all GHA expression used directly in the shell script:
```suggestion
        post: cd %GITHUB_ACTION_PATH% && julia handle_caches.jl rm '${{ github.repository }}' '${{ steps.keys.outputs.restore-key }}' '${{ github.ref }}' '${{ inputs.delete-old-caches != 'required' }}'
```

omus:
> opens the door to arbitrary code execution just by tweaking the workflow name.

If we're concerned about this then I don't think even single quotes will work as an attacker could escape the single quotes as well:
```bash
$ echo ''$(curl -fsSL httpbin.org/headers)''
{ "headers": { "Accept": "*/*", "Host": "httpbin.org", "User-Agent": "curl/8.4.0", "X-Amzn-Trace-Id": "Root=1-67253d14-2e9e98a86356666d2b9009ff" } }
```
Assigning inputs to `env` first (either globally or per step) should avoid this.

DilumAluthge:
Now that we've migrated from composition action to TypeScript action, can I close this PR?

PR #204, opened 2026-03-17
Add Google Cloud Storage cache support

wsmoses:
Adds Google Cloud Storage support for the cache action, including the pidfile exclusion fix.

CC: @gbaraldi

🤖 Generated by AI coding tools.


wsmoses:
gentle ping here for review!

DilumAluthge:
I tried out Codex on the PR, and here's one thought: You're adding a dependency on `@google-cloud/storage`. Is that available in a public NPM registry?

giordano:
It worked in other PRs yesterday though. Does this have to do with forks?

DilumAluthge:
> Ok, now I see it. Is there a way to resolve this issue early then? I see there's an env var called `NPM_CONFIG_REGISTRY` which should configure the registry

Sure, that seems fine too.

gbaraldi:
Why is this a 200k line PR

IanButterworth:
Driveby comment.. I seem to remember that we used to not check `dist/` in, but used some release process.. maybe that'd be better.

DilumAluthge:
> Driveby comment.. I seem to remember that we used to not check `dist/` in, but used some release process.. maybe that'd be better.

It's subjective, but my personal opinion is that it's nice to check `dist/` in:
1. The contents of the PR are exactly the same as what will be released and deployed. So if CI passes on the PR, it will (usually) pass on the release. In contrast, with the old release process, what you test in the PR isn't guaranteed to match the release you build.
2. With dist checked in, making a release is easy, just make the release in GitHub, and update the v3 courtesy tag. In contrast, I personally found the old release process (rhe one we used to have in setup-julia) to be more cumbersome.
3. There's precedent in other actions for checking dist in. For example: https://github.com/actions/checkout/tree/main/dist

giordano (dismissed):
We've been using this PR in https://github.com/EnzymeAD/Enzyme-JAX for a few days now, and it's been working well, the cache is saved to GCP as desired, and not to GitHub Actions.

Does anyone else want to have a look?  Otherwise I'll merge in a few hours.

src/main.js
```diff
@@ -3,6 +3,8 @@ const exec = require('@actions/exec');
 const cache = require('@actions/cache');
 const fs = require('fs');
 const os = require('os');
+const path = require('path');
+const { GoogleCloudStorage } = require('@google-cloud/storage');
```
giordano: I may have broken something with 57daa2cf9de084f639ec6162afb9a37b7cba818d, now we get `Warning: Failed to restore cache from GCS: GoogleCloudStorage is not a constructor`.  Can someone fix it?  I don't have npm on this machine.

DilumAluthge: You're basically trying to do an "import Foo as Bar", right? I'm on mobile, but the JS syntax is something like either `const { GoogleCloudStorage : Storage }` or `const { Storage : GoogleCloudStorage }`.

I agree the rename you want is a good idea, because it makes the code easier to read.

wsmoses:
gentle ping here

cc @giordano

DilumAluthge:
@giordano @IanButterworth Anything else on your end?

Looks like merge conflicts need to be fixed.

DilumAluthge:
Let me do one last review of this, to refresh my memory, since it's been a little while since I looked at it.

handle_caches.jl
```diff
@@ -1,6 +1,10 @@
 using Pkg, Dates
 function handle_caches()
     subcommand = ARGS[1]
+    if isnothing(Sys.which("gh")) || isempty(Sys.which("gh"))
+        @info "GitHub CLI (`gh`) not found. Skipping GitHub Actions cache management."
+        return
+    end
```
DilumAluthge, line 7: Should this throw an error instead?

DilumAluthge (commented):
I have one question that I left as an inline comment.

Additionally, am I remembering correctly that you tried this PR branch out on one of your repos? Did everything work correctly there?

handle_caches.jl
DilumAluthge: How about we do this:
- If the user sets `gcp-bucket`, then skip this check entirely, and just `return nothing` immediately.
- If the user does not set `gcp-bucket`, then throw an error if this check fails.

DilumAluthge:
I think this is ready to merge pending the following:
- [x] Resolve the above discussion about throwing an error if `gh` is not found.
- [x] Confirmation from @wsmoses or @giordano that this PR branch was tested out in a repo, and that everything worked correctly when testing.
    - [x] Done per [this comment](https://github.com/julia-actions/cache/pull/204#issuecomment-4357332781)

handle_caches.jl
giordano: I don't see why this should hard fail. gh may not be present on self-hosted runners, the previous cache isn't deleted, but that's not the end of the world, failing the entire job for this would BE a very bad user experience IMO.

giordano:
Cache is saved and restored correctly in https://github.com/EnzymeAD/Enzyme-JAX/actions/workflows/test-gb-25.yml

handle_caches.jl
DilumAluthge: The result then would be silently failing to delete the previous caches, which means that the repo will quickly hit its total cache size limit, which GitHub is enforcing more aggressively. I don't think we should be silent in such a situation.

Also, prior to this PR, the absence of `gh` would already hard-fail (you'll hit line 12, and then get an error because `gh` is not found). I'm proposing to preserve the status quo in this PR. We could open a separate PR to discuss if we want to change the behavior, but IMO we shouldn't change the behavior in a PR where the main goal of the PR (GCP) is completely unrelated.

giordano: I never had hard errors before. As I said above, on self-hosted runners, or when using custom images, gh may not be available and there's currently no failure. I'm speaking from experience.

DilumAluthge: Ah, you're right. All the calls to `gh` are inside a try-catch, so it doesn't hard error.

I don't love that behavior, but it's the status quo on master, so this PR isn't changing anything. So let's leave this code in (it basically just gives a nicer warning).

DilumAluthge (approved):
LGTM.

DilumAluthge:
Will need a new release.

DilumAluthge:
Should probably bump the minor version number, since this is a new feature.

PR #209, opened 2026-03-20
Delete pidfiles. Skip Pkg.gc when not saving. Don't delete old caches if no new saved. 

### LINKED ISSUE #168: Don't delete caches if new cache wasn't made
While https://github.com/julia-actions/cache/issues/79 remains open we shouldn't be deleting caches if the main run fails. It assumes that a new cache was made.

### LINKED ISSUE #175: Don't Pkg.gc if not going to save cache
Image

### LINKED ISSUE #179: Remove pidfiles from cache
Stale pidfile are automatically deleted by the package manager, but might be cleaner not to store them in the cache in the first place?  Spotted at https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/16999045099/job/48196660878?pr=1307#step:15:293.

IanButterworth:
This PR addresses three related issues in the post action:

- **Exclude pidfiles from cache** (#179): Adds targeted glob exclusions so stale pidfiles created by Julia and Pkg are never stored in or restored from the cache. Patterns are scoped to the exact depot subdirectories where Julia/Pkg place them (`artifacts/`, `compiled/v*.*/`, `packages/`, `registries/`, `logs/`), avoiding accidental exclusion of `.pid`/`.pidfile` files that may be legitimate content inside artifacts or packages.

- **Skip Pkg.gc on exact cache hit** (#175): When the restore key matches the save key exactly, the cache is already up to date and nothing new will be saved. The post action now returns early in this case, avoiding a needless `Pkg.gc()` run.

- **Don't delete old caches if no new cache was saved** (#168): Old-cache deletion now only runs when `saveCache` actually succeeded. If saving fails or is skipped for any reason, the post action returns early so existing caches are left intact.

Fixes #179
Fixes #175
Fixes #168

🤖 Generated by GitHub Copilot (Claude Sonnet 4.6).

### DIFF
--- a/dist/main/index.js
+++ b/dist/main/index.js
@@ -87235,6 +87235,16 @@ async function run() {
         if (cacheScratchspaces) cachePaths.push(scratchspacesPath);
         if (cacheLogs) cachePaths.push(logsPath);
 
+        // Exclude stale pidfiles – they are auto-cleaned but should not be cached.
+        // Each pattern targets only the specific depth where Julia/Pkg places them.
+        // Both .pid and .pidfile extensions are matched for forward-compatibility.
+        cachePaths.push(`!${depotPath}/artifacts/*.{pid,pidfile}`);           // Pkg artifact locks
+        cachePaths.push(`!${depotPath}/compiled/v*.*/*.{pid,pidfile}`);       // Julia base precompile locks (UUID-less packages)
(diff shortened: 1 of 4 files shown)


giordano (approved):

IanButterworth:
Fixed in https://github.com/julia-actions/cache/releases/tag/v3.0.2

PR #222, opened 2026-05-01
Migrate cache cleanup from the `gh` CLI to the `@octokit/rest` NodeJS package

### LINKED ISSUE #221: Migrate from `gh` to a NodeJS package?
Right now, on `main`, we use the `gh` CLI tool. For example:

https://github.com/julia-actions/cache/blob/893200563f9b068e4ef26d680a8c07c93f99a94a/handle_caches.jl#L8-L8

In my opinion, this has some disadvantages:
1. It requires that the user has `gh` available on their CI runner (which might be a self-hosted runner).
2. It adds an external dependency that is not tracked in this repo's `package.json`/`package-lock.json` files, which can make it harder for users to fully track the supply chain of the actions that they use.

Instead of using `gh`, can we use the Octokit NodeJS package? This would fix both of the above issues.

DilumAluthge-LLM:
## Summary

- cherry-pick the `AGENTS.md` commit separation rules from #224
- replace the `gh`-based cache cleanup helper with Octokit REST API calls in the post action
- remove `handle_caches.jl` and the `GH_TOKEN` subprocess path
- keep `package-lock.json` and `dist/*` changes in dedicated commits, rebuilding both bundled actions
- accept DilumAluthge's review suggestions for the Julia `Pkg.gc()` snippet
- keep `@google-cloud/storage` installed because the existing `gcp-bucket` path still imports it

Fixes #221.

## Validation

- `npm install`
- `npm run build`
- `node --check src/main.js && node --check src/post.js`
- `npm run build && test -z "$(git status --porcelain dist/)"`

Not run: `npm run lint` because this PR does not add ESLint; the lint setup is handled in #223.

CC: @DilumAluthge

🤖 Generated by OpenAI Codex.


src/post.js
```diff
@@ -1,10 +1,120 @@
...
-import { Storage as GoogleCloudStorage } from '@google-cloud/storage';
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
+import { Octokit } from '@octokit/rest';
+
+async function runPkgGc() {
+    const exitCode = await exec.exec('julia', [
+        '-e',
+        [
+            'try',
+            '    using Pkg',
```
DilumAluthge:
```suggestion
            '    import Pkg',
```

src/post.js
```diff
@@ -1,10 +1,120 @@
...
+            '    Pkg.gc()',
+            'catch e',
+            '    @error "An error occurred while managing existing caches" e',
+            '    exit(17)',
+            'end'
```
DilumAluthge:
```suggestion
            'catch',
            '    error("An error occurred while managing existing caches")',
            'end'
```

PR #244, opened 2026-07-23
Add option to use (multithreaded) zstd for gcp

wsmoses:
Speed up the compress/unzip 

@giordano @DilumAluthge

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -18,6 +18,9 @@ inputs:
   gcp-bucket:
     description: Google Cloud Storage bucket to use for caching (e.g., 'my-cache-bucket'). If provided, this is used instead of the default Actions cache.
     default: ''
+  gcp-compression:
+    description: Compression algorithm to use for GCP storage caching ('zstd' or 'gzip').
+    default: 'zstd'
   depot:
     description: Path to a Julia depot directory where cached data will be saved to and restored from.
     default: ''
--- a/src/main.js
+++ b/src/main.js
@@ -4,12 +4,35 @@ import * as cache from '@actions/cache';
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
-import { spawn } from 'child_process';
+import { spawn, execSync } from 'child_process';
 import { Storage as GoogleCloudStorage } from '@google-cloud/storage';
 
-function streamRestore({ inStream, cwd }) {
+function isZstdAvailable() {
+    try {
+        execSync('zstd --version', { stdio: 'ignore' });
+        return true;
+    } catch {
+        return false;
+    }
+}
+
+function parseGcpCompressionInput(inputVal) {
+    if (!inputVal || inputVal.trim() === '') {
+        return 'zstd';
+    }
+    const val = inputVal.trim().toLowerCase();
+    if (val === 'zstd' || val === 'gzip') {
+        return val;
+    }
+    throw new Error(`Invalid compression value for input 'gcp-compression': '${inputVal}'. Expected 'zstd' or 'gzip'.`);
+}
+
+function streamRestore({ inStream, useZstd, cwd }) {
     return new Promise((resolve, reject) => {
-        const decompressProc = spawn('gzip', ['-d', '-c'], { stdio: ['pipe', 'pipe', 'inherit'] });
+        const decompressCmd = useZstd ? 'zstd' : 'gzip';
+        const decompressArgs = useZstd ? ['-d', '-c'] : ['-d', '-c'];
+        const decompressProc = spawn(decompressCmd, decompressArgs, { stdio: ['pipe', 'pipe', 'inherit'] });
+
         const tarProc = spawn('tar', ['-xf', '-'], { cwd, stdio: ['pipe', 'inherit', 'inherit'] });
 
         let errorOccurred = false;
@@ -29,7 +52,7 @@ function streamRestore({ inStream, cwd }) {
 
         decompressProc.on('close', (code) => {
             if (code !== 0 && !errorOccurred) {
-                onError(new Error(`gzip process failed with exit code ${code}`));
+                onError(new Error(`${decompressCmd} process failed with exit code ${code}`));
             }
         });
 
@@ -62,6 +85,11 @@ async function run() {
         const token = core.getInput('token');
         const saveAlways = core.getInput('save-always') === 'true';
         const gcpBucket = core.getInput('gcp-bucket');
+        const gcpCompression = parseGcpCompressionInput(core.getInput('gcp-compression'));
+
+        if (gcpBucket && gcpCompression === 'zstd' && !isZstdAvailable()) {
+            throw new Error("zstd is not available on this runner. Please install zstd or set `gcp-compression: 'gzip'` to use gzip compression.");
+        }
 
         // Determine depot path
         let depotPath;
@@ -188,6 +216,7 @@ async function run() {
         core.saveState('ref', ref);
         core.saveState('default-branch', defaultBranch);
         core.saveState('gcp-bucket', gcpBucket);
+        core.saveState('gcp-compression', gcpCompression);
 
         // Restore cache
         let cacheHit = '';
@@ -196,9 +225,11 @@ async function run() {
                 try {
                     const storage = new GoogleCloudStorage();
                     const bucket = storage.bucket(gcpBucket);
+                    const useZstd = gcpCompression === 'zstd';
+                    const ext = useZstd ? '.tar.zst' : '.tar.gz';
                     let restoredKey = '';
 
-                    const exactFile = bucket.file(`${key}.tar.gz`);
+                    const exactFile = bucket.file(`${key}${ext}`);
                     const [exactExists] = await exactFile.exists();
 
                     let fileToStream = null;
@@ -206,7 +237,7 @@ async function run() {
                         fileToStream = exactFile;
                         restoredKey = key;
                     } else {
-                        const restoreFile = bucket.file(`${restoreKey}.tar.gz`);
+                        const restoreFile = bucket.file(`${restoreKey}${ext}`);
                         const [restoreExists] = await restoreFile.exists();
                         if (restoreExists) {
                             fileToStream = restoreFile;
@@ -220,7 +251,7 @@ async function run() {
                         core.saveState('cache-matched-key', restoredKey);
                         const cwd = process.platform === 'win32' ? depotPath.split(':')[0] + ':/' : '/';
                         const inStream = fileToStream.createReadStream();
-                        await streamRestore({ inStream, cwd });
+                        await streamRestore({ inStream, useZstd, cwd });
                     } else {
                         core.info('No cache found in GCS');
                     }
--- a/src/post.js
+++ b/src/post.js
@@ -2,16 +2,26 @@ import * as core from '@actions/core';
 import * as exec from '@actions/exec';
 import * as cache from '@actions/cache';
 import path from 'path';
-import { spawn } from 'child_process';
+import { spawn, execSync } from 'child_process';
 import { fileURLToPath } from 'url';
 import { Storage as GoogleCloudStorage } from '@google-cloud/storage';
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
-function streamSave({ outStream, cwd, excludePaths, includePaths }) {
+function isZstdAvailable() {
+    try {
+        execSync('zstd --version', { stdio: 'ignore' });
+        return true;
+    } catch {
+        return false;
+    }
+}
+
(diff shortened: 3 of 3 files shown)


DilumAluthge:
Can you put the changes that touch `dist/*` into a separate commit?

wsmoses:
done

DilumAluthge:
Instead of implementing the zstd-calling ourselves, could we use an existing NodeJS package? E.g. something like https://www.npmjs.com/package/simple-zstd?

---

(I was originally going to suggest using `node:zlib`, but it looks like the zstd support in `node:zlib` is "stability 1 - experimental", so maybe not ready for us to use here.)

wsmoses:
If we're already using the system installed one I don't feel particularly strongly about the wrapper package (since we also need to do the existence check for the binary ourselves if it doesn't exist)

DilumAluthge:
If using a wrapper package can reduce the number of lines of code that we have to maintain, then I think it's worth trying. Can you give it a try, and see if it reduces the amount of code we have?

wsmoses:
I don't think it's legal for us to use it per:

```
zstd: Must be installed and available on system PATH
```

which cannot be assumed [and we check this manually ourselves here, falling back if not]

DilumAluthge:
I was thinking that we would import simple-zstd, but not actually call any of its functions until we first make sure that zstd is in the PATH.

wsmoses:
Doesn't that rely on that package not using the binary during any initialization though?

DilumAluthge:
Yeah, you're right. I just looked through the source code (https://github.com/Stieneee/simple-zstd/blob/main/src/index.ts), and it does seem to eagerly search for zstd in the path.

Instead, how about:
1. We check if zstd is in the PATH
2. We dynamically import `simple-zstd`, but only if zstd is in the PATH.

GPT wrote this example snippet for me:

```
const simpleZstd = isZstdAvailable()
  ? await import("simple-zstd")
  : undefined;
```

wsmoses:
I think that will only add lines of code since we'd need here though? And wed still need the manual zstd call to check if it exists...

src/post.js
```diff
@@ -2,10 +2,65 @@ import * as core from '@actions/core';
...
+function isZstdAvailable() {
+    try {
+        execSync('zstd --version', { stdio: 'ignore' });
+        return true;
+    } catch (e) {
+        return false;
+    }
+}
+
+function streamSave({ outStream, useZstd, cwd, excludePaths, includePaths }) {
```
DilumAluthge: So, my hope was that using simple-zstd (or some other external package) would reduce (or eliminate) some of the lines of code from functions like `streamSave` here.

Would that not be the case?

wsmoses:
No because if you don't use zstd you still need most of the code for tar.

Sorry this PR also further implements an optimization on the stream itself

wsmoses:
Specifically it avoids writing a temporary file which is uploaded and instead directly streams the compressed result to the network for upload/download

DilumAluthge:
Ah, okay that's probably why the diff is so big.

Could you split it into two PRs? One for the no-intermediate-file change, and one for the zstd change? You can stack one PR on top of the other (which you'll need commit access to do - I've invited you).

Also, if it helps, you can hold off on pushing the `dist/` changes initially, and just push them once we've finished review.

wsmoses:
so I'm locally testing these things [so the dist is useful]

wsmoses:
@DilumAluthge this is the second one [cc @giordano ]

action.yml
```diff
@@ -18,6 +18,9 @@ inputs:
   gcp-bucket:
     description: Google Cloud Storage bucket to use for caching (e.g., 'my-cache-bucket'). If provided, this is used instead of the default Actions cache.
     default: ''
+  gcp-zstd:
```
DilumAluthge: Instead of making this toggleable by the user, should we just check if `zstd` is in the PATH, and then use it if it is found?

wsmoses: is there a reason to not use zstd (e.g. is it potentially larger/may have an issue/etc, maybe a different job/container didn't have it installed and wanted to make sure everything went to the same cache)? if so a toggle may be useful.

wsmoses:
@DilumAluthge now rebased off main

action.yml
DilumAluthge: > maybe a different job/container didn't have it installed and wanted to make sure everything went to the same cache

Yeah, this one would be problematic. If job 1 uploads a new cache with `.tar.zst`, then if job 2 tries to download the `.tar.zst` cache but job 2 doesn't have `zstd` available in PATH, then that would be problem.

Idk how common that use case would be? Because if you want jobs 1 and 2 to share the same cache, presumably you also want to run them in similar enough environments/containers?

wsmoses: I can see that, but also that's why I figure it's better just to leave it as an option.

DilumAluthge: Yeah, I think we have to leave it as an option.

Does regular GitHub Actions caching (non-GCP) support `zstd`? If so, should we rename the option to just `use-zstd` or something, instead of being `gcp-zsdt`?

src/main.js
```diff
@@ -4,12 +4,24 @@ import * as cache from '@actions/cache';
...
-import { spawn } from 'child_process';
+import { spawn, execSync } from 'child_process';
 import { Storage as GoogleCloudStorage } from '@google-cloud/storage';
 
-function streamRestore({ inStream, cwd }) {
+function isZstdAvailable() {
+    try {
+        execSync('zstd --version', { stdio: 'ignore' });
+        return true;
+    } catch (e) {
```
DilumAluthge: I think this should work:

```suggestion
    } catch {
```

src/main.js
```diff
@@ -196,31 +210,52 @@ async function run() {
...
+                            { key: key, ext: '.tar.zst', useZstd: true }
+                        ]),
+                        ...(zstdAvailable ? [
+                            { key: restoreKey, ext: '.tar.zst', useZstd: true },
+                            { key: restoreKey, ext: '.tar.gz', useZstd: false }
+                        ] : [
+                            { key: restoreKey, ext: '.tar.gz', useZstd: false },
+                            { key: restoreKey, ext: '.tar.zst', useZstd: true }
+                        ])
+                    ];
```
DilumAluthge: I'm not sure I understand this part. If `zstdAvailable` is false, it looks like it still tries to download `.tar.zst` caches if available? That seems incorrect?

src/post.js
DilumAluthge: This was discussed elsewhere. IIRC, the conclusion was that using an external package (like simple-zstd) wouldn't significantly reduce the number of lines of code we would need to maintain here.

src/post.js
```diff
@@ -63,6 +75,7 @@ async function run() {
         const ref = core.getState('ref');
         const defaultBranch = core.getState('default-branch');
         const gcpBucket = core.getState('gcp-bucket');
+        const gcpZstd = core.getState('gcp-zstd') !== 'false';
```
DilumAluthge: Seems like we should try harder to parse the string as a bool, and throw an error if the string isn't `true` or `1` or `false` or `0`?

And you probably want to throw the error earlier (in `src/main.js`), not in the post step.

src/post.js
```diff
@@ -2,16 +2,28 @@ import * as core from '@actions/core';
...
+}
+
+function streamSave({ outStream, useZstd, cwd, excludePaths, includePaths }) {
     return new Promise((resolve, reject) => {
         const tarArgs = ['-cf', '-', ...excludePaths, ...includePaths];
         const tarProc = spawn('tar', tarArgs, { cwd, stdio: ['ignore', 'pipe', 'inherit'] });
-        const compressProc = spawn('gzip', ['-c'], { stdio: ['pipe', 'pipe', 'inherit'] });
+
+        const compressCmd = useZstd ? 'zstd' : 'gzip';
+        const compressArgs = useZstd ? ['-T0'] : ['-c'];
```
DilumAluthge: Is this code redundant with lines 123-124 below?

DilumAluthge (changes requested):

action.yml
DilumAluthge: Oh, also, instead of being zstd yes/no, can we instead make it something like:

```yaml
uses: julia-actions/cache
  with:
    compression: zstd,gzip
```

That is, a comma-separated list of acceptable compression algorithms? That way, we can easily add others in the future?

Default value would be `compression: '*'`, which would mean "all algorithms allowed".

We could name it `gcp-compression` if it's definitely GCP-specific, but it would be nice if you could figure out whether regular (non-GCP) GitHub Actions caching supports zstd or any other non-gzip algorithms for compression/decompression.

wsmoses: I think it depends on the container setup as to whether not that exists. It doesn't make sense here to use one for a global zstd though since the normal path just calls

```
import * as cache from '@actions/cache';
...
const restoredKey = await cache.restoreCache(cachePaths, key, [restoreKey]);
```

so we don't have control over that

DilumAluthge: Ah, yeah, if `cache.restoreCache` doesn't accept a compression algorithm as an argument, then I think the option needs to be GCP-specific. How about this?

```yaml
uses: julia-actions/cache
  with:
    gcp-compression: zstd,gzip
```

wsmoses: I actually think we might want to go the other way [and rewrote it a bit], specifically to exclusively use one form of compression. aka if it uses zstd it won't look at gzip caches, etc. I think that actually is the correct way of doing things [to avoid getting stale or conflicting caches from other jobs]

DilumAluthge: Hmmm, so you don't anticipate adding any other compression algorithm options in the future?

My motivation here is to avoid ending up with multiple inputs of the form `gcp-zstd`, `gcp-xz`, etc in the future.

src/main.js
```diff
@@ -62,6 +88,11 @@ async function run() {
         const token = core.getInput('token');
         const saveAlways = core.getInput('save-always') === 'true';
         const gcpBucket = core.getInput('gcp-bucket');
+        const gcpZstd = parseBoolInput(core.getInput('gcp-zstd'), 'gcp-zstd', true);
```
DilumAluthge: Do you need the default value of `true` here? IIRC, by specifying `default: true` in `action.yml`, you'll already get the default value, right?

src/main.js
```diff
@@ -62,6 +88,11 @@ async function run() {
...
+
+        if (gcpBucket && gcpZstd && !isZstdAvailable()) {
+            throw new Error("zstd is not available on this runner. Please install zstd or set `gcp-zstd: 'false'` to use gzip/tar compression.");
```
DilumAluthge: It's always saved as a tarball file in the GCP bucket, right? Regardless of whether it's `tar.gz` or `tar.zst`?
```suggestion
            throw new Error("zstd is not available on this runner. Please install zstd or set `gcp-zstd: 'false'` to use gzip compression.");
```

src/post.js
DilumAluthge:
```suggestion
        const gcpZstd = core.getState('gcp-zstd') == 'true'; // Argument validation was performed in [function name] in `src/main.js`
```

DilumAluthge (changes requested):

wsmoses:
okay changed it to take an algorithm, resolving most of above comments

src/main.js
```diff
@@ -4,12 +4,35 @@ import * as cache from '@actions/cache';
...
+        return true;
+    } catch {
+        return false;
+    }
+}
+
+function parseGcpCompressionInput(inputVal) {
+    if (!inputVal || inputVal.trim() === '') {
+        return 'zstd';
+    }
```
DilumAluthge, line 22: I don't think this is strictly necessary (because the default is already specified in `action.yml`), but technically it doesn't hurt to keep (it just duplicates the default value in two places).

src/post.js
```diff
@@ -63,6 +73,8 @@ async function run() {
         const ref = core.getState('ref');
         const defaultBranch = core.getState('default-branch');
         const gcpBucket = core.getState('gcp-bucket');
+        const gcpCompression = core.getState('gcp-compression') || 'zstd';
```
DilumAluthge: I'm still confused by `|| 'zstd'`. In what situation would the `gcp-compression` state value not have been set in `src/main.js`? The other `core.getState` calls don't seem to have default values?

DilumAluthge (commented):
1. One comment (that doesn't need to be addressed, just an observation)
2. One question

After the one question is resolved, this LGTM to merge.

src/post.js
wsmoses: yeah I'm not sure, but can remove;

DilumAluthge (approved):

DilumAluthge:
@wsmoses Can I merge this one once CI passes?

wsmoses:
Yeah

PR #245, opened 2026-07-24
Stream tar archives directly to/from GCP storage without intermediate files

wsmoses:


wsmoses:
@DilumAluthge this is the first one [cc @giordano ]

src/main.js
```diff
@@ -4,8 +4,48 @@ import * as cache from '@actions/cache';
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
+import { spawn } from 'child_process';
 import { Storage as GoogleCloudStorage } from '@google-cloud/storage';
 
+function streamRestore({ inStream, cwd }) {
+    return new Promise((resolve, reject) => {
+        const decompressProc = spawn('gzip', ['-d', '-c'], { stdio: ['pipe', 'pipe', 'inherit'] });
```
DilumAluthge, line 12: Just a note to my self: This isn't adding a new requirement. The old code already called `tar -zcf`, which requires that `gzip` is on the PATH, so we already required `gzip`.

src/main.js
```diff
@@ -4,8 +4,48 @@ import * as cache from '@actions/cache';
...
+            }
+        };
+
+        inStream.on('error', onError);
+        decompressProc.on('error', onError);
+        tarProc.on('error', onError);
+
+        decompressProc.on('close', (code) => {
+            if (code !== 0 && !errorOccurred) {
+                onError(new Error(`gzip process failed with exit code ${code}`));
```
DilumAluthge, line 32: Would there be any benefit to printing the `gzip` process's stderr to the CI logs if a failure occurs?

Same comment for the other places where we check the exit code of the `gzip` or `tar` process.

Note: This doesn't have to block merging of this PR - I think it's a potential logging improvement that could happen in a follow-up PR, if we think it would be useful.

DilumAluthge (approved):

DilumAluthge:
LGTM.

@wsmoses Can I merge this one?

src/main.js
wsmoses: maybe, but offhand I'm not sure what the error would be and if it would be actionable? happy to ponder in post

wsmoses:
go for it!

PR #246, opened 2026-07-24
Add optional prefix for uploads

wsmoses:
Adding a prefix is especially relevant for gcp-style uploads, in particular unlike github where all caches are separated per repo, a single bucket may be used for multiple repos, requiring a prefix to avoid accidentally having two different repos sharing the same bucket from sharing the same cache files

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -21,6 +21,9 @@ inputs:
   gcp-compression:
     description: Compression algorithm to use for GCP storage caching ('zstd' or 'gzip').
     default: 'zstd'
+  key-prefix:
+    description: Optional prefix to prepend to the cache key (and GCS object path, if using GCP).
+    default: ''
   depot:
     description: Path to a Julia depot directory where cached data will be saved to and restored from.
     default: ''
--- a/src/main.js
+++ b/src/main.js
@@ -86,6 +86,7 @@ async function run() {
         const saveAlways = core.getInput('save-always') === 'true';
         const gcpBucket = core.getInput('gcp-bucket');
         const gcpCompression = parseGcpCompressionInput(core.getInput('gcp-compression'));
+        const keyPrefix = core.getInput('key-prefix');
 
         if (gcpBucket && gcpCompression === 'zstd' && !isZstdAvailable()) {
             throw new Error("zstd is not available on this runner. Please install zstd or set `gcp-compression: 'gzip'` to use gzip compression.");
@@ -187,7 +188,7 @@ async function run() {
             }
         }
 
-        let restoreKey = `${cacheName};os=${runnerOS};${matrixKey}`;
+        let restoreKey = `${keyPrefix}${cacheName};os=${runnerOS};${matrixKey}`;
         // URL encode restricted characters
         restoreKey = restoreKey.replace(/,/g, '%2C');
 


wsmoses:
@DilumAluthge this is the third and final one [cc @giordano ]

action.yml
```diff
@@ -18,6 +18,12 @@ inputs:
   gcp-bucket:
     description: Google Cloud Storage bucket to use for caching (e.g., 'my-cache-bucket'). If provided, this is used instead of the default Actions cache.
     default: ''
+  gcp-zstd:
+    description: Enable zstd compression for GCP storage caching if zstd is available on the runner.
+    default: 'true'
+  gcp-prefix:
```
DilumAluthge: Does this need to be GCP specific? Could we instead just let people provide an arbitrary cache key prefix?

wsmoses: done

action.yml
```diff
@@ -21,6 +21,9 @@ inputs:
   gcp-zstd:
     description: Enable zstd compression for GCP storage caching if zstd is available on the runner.
     default: 'true'
+  prefix:
+    description: Optional prefix to prepend to the cache key (and GCS object path).
```
DilumAluthge: This isn't GCP-specific, right?

wsmoses: rewrite it to not be gcp specific [it is now generic for the key]

wsmoses: rewrote***

DilumAluthge:
```suggestion
    description: Optional prefix to prepend to the cache key (and GCS object path, if using GCP).
```

DilumAluthge:
Can you update the PR title, since this is no longer GCP-specific?

Also, any chance you could stick a brief background or motivation in the PR description, just briefly outlining the use case for this?

action.yml
```diff
@@ -18,6 +18,12 @@ inputs:
   gcp-bucket:
     description: Google Cloud Storage bucket to use for caching (e.g., 'my-cache-bucket'). If provided, this is used instead of the default Actions cache.
     default: ''
+  gcp-compression:
+    description: Compression algorithm to use for GCP storage caching ('zstd' or 'gzip').
+    default: 'zstd'
+  prefix:
```
DilumAluthge: This is a bit of bikeshedding, but maybe something like `key-prefix` would be more meaningful as a name?

```suggestion
  key-prefix:
```

wsmoses:
done

action.yml
wsmoses: ehhhh, I feel like prefix is clear enough here, but I am amenable if you insist. It's also equivalent the prefix of the file as the key

DilumAluthge: I don't feel too strongly. Since it is going to be new public API (and thus can't be changed later without a breaking release), maybe we can do a straw poll?

@IanButterworth @omus @giordano Any opinions on the naming of this input? It's a string that will be prepended to the cache key.

wsmoses: resolved per slack as key-prefix

wsmoses:
now rebased to main

src/main.js
```diff
@@ -86,6 +86,7 @@ async function run() {
         const saveAlways = core.getInput('save-always') === 'true';
         const gcpBucket = core.getInput('gcp-bucket');
         const gcpCompression = parseGcpCompressionInput(core.getInput('gcp-compression'));
+        const prefix = core.getInput('key-prefix');
```
DilumAluthge:
```suggestion
        const keyPrefix = core.getInput('key-prefix');
```

src/main.js
```diff
@@ -187,7 +188,7 @@ async function run() {
             }
         }
 
-        let restoreKey = `${cacheName};os=${runnerOS};${matrixKey}`;
+        let restoreKey = `${prefix}${cacheName};os=${runnerOS};${matrixKey}`;
```
DilumAluthge:
```suggestion
        let restoreKey = `${keyPrefix}${cacheName};os=${runnerOS};${matrixKey}`;
```

wsmoses: done

DilumAluthge (approved):
LGTM.

PR #253, opened 2026-07-28
`.gitattributes`: Hide generated JS code from diffs and language stats

DilumAluthge:
Co-authored-by: Sascha Mann @SaschaMann


DilumAluthge:
cc: @SaschaMann

PR #254, opened 2026-07-31
fix: upgrade brace-expansion to 5.0.7, 1.1.16, 2.1.2 (CVE-2026-13149)

anupamme:
## Summary
Upgrade brace-expansion from 1.1.13 to 5.0.7, 1.1.16, 2.1.2 to fix CVE-2026-13149.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | CVE-2026-13149 |
| **Severity** | HIGH |
| **Scanner** | trivy |
| **Rule** | `CVE-2026-13149` |
| **File** | `package-lock.json` |
| **Assessment** | Likely exploitable |

**Description**: brace-expansion: Brace-expansion: Denial of Service due to exponential-time complexity

## Evidence

**Scanner confirmation**: trivy rule `CVE-2026-13149` flagged this pattern.

**Production code**: This file is in the production codebase, not test-only code.

## Changes
- `package.json`
- `package-lock.json`

## Behavior Preservation
The change is scoped to 2 files on the vulnerable path, and the project builds successfully with this change applied.

## Verification
- [x] Build passes
- [x] Scanner re-scan confirms fix
- [x] LLM code review passed

---
*This change addresses a pattern flagged by static analysis. The code path handles user-influenced input and the fix reduces the attack surface against both manual and automated exploitation.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -581,15 +581,6 @@
         "node": "*"
       }
     },
-    "node_modules/brace-expansion": {
-      "version": "2.1.2",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.2.tgz",
-      "integrity": "sha512-w5JZcKgdhDOgOwm8H+KgbosopHMuGcl6qbulwjtz3SM7I7P3yW1eAjzMPLrIE+NQ9vjgANKHWeMHnrT0OXW1oA==",
-      "license": "MIT",
-      "dependencies": {
-        "balanced-match": "^1.0.0"
-      }
-    },
     "node_modules/buffer-equal-constant-time": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/buffer-equal-constant-time/-/buffer-equal-constant-time-1.0.1.tgz",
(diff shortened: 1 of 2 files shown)


DilumAluthge:
Thank you for the PR. I notice you're doing an `override` here. I'm not necessarily opposed to using an `override`. However, ideally, the issue would be fixed in the upstream packages that depend directly on `brace-expansion`.

anupamme:
Good catch, and agreed. I dug into the tree; the only thing pulling in brace-expansion is @actions/glob → minimatch@3.1.5, which already declares brace-expansion: ^1.1.7. That range happens to include 1.1.16, the patched 1.x release for this CVE, so I don't actually need the overrides block or the jump to 2.1.2; I can update the lockfile so brace-expansion resolves to 1.1.16 within minimatch's own existing semver range. That's a more minimal, in-range fix and avoids forcing a major-version bump on a dependency minimatch wasn't tested against. I'll push that update. Longer-term, the real fix is upstream in @actions/toolkit's @actions/glob, which is still pinned to minimatch@3.x.

DilumAluthge:
I think this PR is no longer needed. Let me know if I am incorrect.

anupamme:
Thanks for checking! Agreed, if the vulnerable dependency has already been resolved on main, there’s no need to keep this PR open.

I initially proposed an override, but after tracing the dependency tree, I updated the PR to use the existing semver range and resolve brace-expansion to 1.1.16 instead.

